PR #143, opened 2015-03-13
Changes to use tabulated vfky cross-sections stored in the phot_top structure

jhmatthews:
Not for merging yet. Some fairly major changes here. This is not as simple as just tabulating the data and going from there, as I've had to change a bunch of different code which used the old vfky routines, and also changed some routines which only looped over the topbase cross-sections. The code will use topbase if it's available and if it has nlte_levels set to > 0. 

Each ion has the following variables relating to Xsections
- phot_info: 0 - vfky, 1- topbase, -1- no xsection (hopefully means it's ionized)
- ntop_ground: where the ground state is in phot_top, if it's topbase
- ntop: number of topbase xsections for this ion
- nxphot: where the vfky xsection is in phot_top, if it's vfky. will always be higher than the total number of topbase xsections.

Summary of changes:
- sigma_phot_topbase is now called sigma_phot. It will calculate the cross section for either a tabulated vfky or topbase Xsection by interpolation
- the old routine sigma_phot, which used the vfky fitting formula, is now gone, and any routines which used it have to be adapted
- the frequency ordered photoionization xs array now contains TopBase and Vfky data
- in radiation.c, we used to loop over ntop_phot, and call sigma_phot_topbase, then loop over vfky and call sigma_phot. Now there's one loop over ntop_phot + nxphot
- in recomb.c there are various changes to stop one using vfky, and the function fb_partial_verner is gone.
- in the macro-atom routines we used to only loop over topbase xsections. I believe we should loop over both when considering kpkt cooling channels and simple ion bf processes.
- total_fb which calculates the non-macro-atom fb cooling was only looping over ntop_phot as well.

This needs some fairly significant testing, as it's not just a tabulation change- there will probably be changes in the fb cooling, especially in matom mode. I will test a version with some of the extra loops removed- i.e. one that I believe should produce the same results as the old code, because the only change should be tabulation. I will then test this version against that, to see what the effect is. 

The tabulated xsections will go in data when I merge this. They are created by running

```
$PYTHON/py_progs/photo_xs.py tab data/atomic77/photo_verner.data output_filename
```


jhmatthews:
Merging this. Test results here: https://sites.google.com/site/agnwinds/bugfixes-and-tests/testing-changes-to-xsections

atomic data repo has been updated. Use standard78 atomic data to use the tabulated xsections. They are stored in the file atomic78/photo_vfky_tabulated.data, and are created using the script py_progs/photo_xs.py, with instructions in the docstring for that script.

PR #145, opened 2015-04-27
Reverb Code Merge

smangham:
Code for reverberation mapping. The post processing code used to build transfer function maps from the .delay_dump files is at:
https://github.com/smangham/bindata

The following option(s) are added to input files after "Use.standard.care.factors(1=yes)":
reverb.type(0=no,1=phot,2=wind)         2
reverb.path_bins                        30 - If type 2, sets number of path bins discretised 
reverb.theta_bins                       30 - If type 2, used for output of .vtk file


jhmatthews:
Hey Sam. Thanks for doing this. Few questions while I remember:
- currently failing the build test- I think it's probably because there's some stuff missing from the py_wind_objects or similar in the Makefile?
- we probably want to update the source files in dev to your new format of type declaratons in functions, if that's what we are adopting, to make the diffs easier to compare - makes it better if we look back at what changed in the future.
- is the variance stuff supposed to be merged? I notice the new functions are tracked in this
- generally "code policy" is to avoid printf and use Log() because then it goes into the diag file too. Not sure how you were supposed to know that though ;)

smangham:
I think it needs time.c and wind_keplerian.c if those haven't automatically merged in- I think wind_keplerian uses time functions somewhere, explaining the time dependency.

Indent should automatically switch the declaration format.

The VR stuff shouldn't be merged in- there's no splitting or anything being called. If there's any VR code in there it's just adding the options to geo (but not even reading them in) and importance fields to the wind. 

I assumed log just printed to the diag files- should I be using it for stuff I want to print to screen as _well_ as log?

jhmatthews:
I guess the way to think about it is if something isn't worth logging to the diagfile then it's not worth having in the proper version of the code - so any messages to user use either Log or Error. See e.g. Knox's writing here: https://github.com/agnwinds/python/wiki/Avoid-printf - it's not something which particularly bothers me but it's very easy to end up clogging the on screen printing. 

VR- I'd suggest the functions aren't tracked in this merge then. Separate branch for VR work?

RE declaration format- happy to change this if needs be but in a separate commit to clean up this diff. Need to check others are OK and should probably discuss indent format while we're at it.

kslong:
HI Sam,
I do feel strongly about this.   We have created a flexible system for logging that allows one to send items to the screen and to the log file.  Only the most important messages should be sent to the screen, and the ones that are sent to the screen should be clear enough that most users can understand them.  If you must use printf statements (rather than a proper debugger), they should be regarded as very temporary, and removed as soon as you have figured out a problem.
Cheers,
Knox

Issue #148, opened 2015-05-18
Style - Enums


smangham:
The code makes a lot of use of integers and lists of integer aliases for states (e.g. wind.inwind and the W_ALL_INWIND etc. list) that are functionally equivalent to enums, but doesn't actually include any enums. 

Is there a policy against using enums, or is this just a case of personal preference? Would anyone be opposed to the use of enums in new code?

Higginbottom:
what's an enum?

smangham:
An enumerated list, they're declared and used like below (where the 0 = default):
<i>enum {VR_NONE=0, VR_IONISATION, VR_SPECTRUM, VR_BOTH_CYCLES} vr_type;</i>
<i>if(vr_type != VR_NONE) [...]
vr_type = VR_BOTH_CYCLES;</i>
On compile they're converted to the sort of int & series of #define ints that we use extensively. They're just neater (currently our #defines can be dozens of lines away from the variable they're associated with) and allow for compiler checks for invalid sets.
You _can_, if wanted, set values for the other enum values (e.g. VR_IONISATION=1) and compare between enums, read ints into them etc. so there's no loss of functionality compared to ints, you'll just get warnings about anything that the compiler can't properly check.

Higginbottom:
Thanks! I guess I have no particular problem with using them, how much of the code needs to be changed? Are the comparisons that crop up everywhere like if(geo.wind_type = proga){ 
still valid, or is there a different syntax? If so, it might be a bit of a drag to go through and change everything.....

smangham:
Well, I wasn't proposing refactoring into enums, just adding new enums instead of new ints/defines in my code. Though all the syntax should be the same so a refactoring would be easy AFAIK- the only bit where the syntax might be an issue is using rdint to fill enums in the get_params section, and even then you'd just get a warning.

kslong:
My question was the same as NIck’s, so I looked it up.

The reason enums are not in PYTHON is that they do not exist in KR c which is what I learned.  Indeed, when PYTHON was initially written we used integers for choices.  I replaced that with the #define statements to make it easer to keep track of what we were actually doing as the code got more and more complex.

Enums may be reasonable now; they are part of ANSI C so it should not matter as a matter of portability.

[...]

PR #158, opened 2015-07-06
Enum

smangham:
This is the switch to enums made for discussion #148. The enum-likes not switched to enum are SPECTYPE, IONMODE and NEBULARMODE. 

Spectype readin is unmodified as the get_spectype/get_models chain is convoluted- switching from int to enum will have no impact on readability or maintenance. IONMODE and NEBULARMODE currently depend on modes sharing integer values and switching to enums would slightly obfusciate that (unless a new comparison function was added).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,8 +21,8 @@ compiler:
 # only want to run this test when we commit to dev
 branches:
   only:
-      dev
-      enum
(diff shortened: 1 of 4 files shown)


jhmatthews:
Has this been tested Sam? Can we merge?

PR #178, opened 2015-09-30
New hydro code

Higginbottom:
This set the groundwork for connecting python to zeus, plus some other housekeeping

### DIFF
--- a/py_progs/hydro_2_python.py
+++ b/py_progs/hydro_2_python.py
@@ -300,7 +300,7 @@ temp=(3.0/2.0)*data["TOTAL ENERGY"]
 data["TEMPERATURE"]=temp/((data["DENSITY"]/(consts.m_p.cgs*MU))*consts.k_B.cgs)
 
 # Open an output file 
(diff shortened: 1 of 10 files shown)


jhmatthews:
@Higginbottom - so the reason for this failing build tests is that it doesn't compile with gcc, but does with mpicc. This is just because the `FILE *fptr, *fopen ();' line is inside an #ifdef so isn't declared without MPI_ON compiler flag (line 115 wind_updates2d.c).

``` C
#ifdef MPI_ON
  int num_mpi_cells, num_mpi_extra, position, ndo, n_mpi, num_comm, n_mpi2;
  int size_of_commbuffer;
  char *commbuffer;
  FILE *fptr, *fopen ();
```

could you fix and push? I presume that file ptr wants to be declared in all cases?

Higginbottom:
Genius!
Thanks James, will fix right now!

jhmatthews:
Cool. I'm merging into dev via command line.

PR #182, opened 2015-10-28
Local domain2

Higginbottom:
Changes by NSH to get hydro import working


jhmatthews:
Nick and Knox @Higginbottom @kslong -- I've now merged Nick's changes detailed in this pull request. I haven't checked the code actually works with hydro but I believe Nick has, and most of the changes here are just to get hydro import working with domains.

PR #189, opened 2015-11-14
Added nioniz to parallel comms

Higginbottom:
I've been just a tiny bit confused and worried by the way that I've been seeing cells in my radhydro models with huge ionization parameters, but apparently very few ionizing photons as a fraction of the total number of photons. Turns out that we were not collecting nioinz across threads....
This pull fixes that!


jhmatthews:
Looks good @Higginbottom , merging :)

jhmatthews:
Had to do a bit of fiddling around with git rebase -- this commit got squashed into https://github.com/agnwinds/python/commit/3c316afc2e8a15003e8041851117e46ed1a06dd5

PR #202, opened 2016-03-07
Multiple bugfixes and improvements

Higginbottom:
There are a few modifications to python in this pull, each in a separate commit (more or less). The headlines are:
1: Changed the value of RADIAN to increased precision. This was to cope with some problems arising in finely gridded hydro models. This gave some differences in an spectrum - but within random photon errors.
2: Modified the bremstrahlung central source spectrum to allow a user defined alpha (the exponent of the nu term) to permit matching either cloudy or xstar. - no change in fiducial AGN spectrum
3: Made a small change to the direct ion calculations. The check to see if coffee were available was pointing at the wrong record. Also, if another check proved incorrect, the error message said it was setting the coeff to zero - but it didn't. Small changes to fiducial AGN, but within random photon noise
4: Small change to photon stat output. We now save the mean weight of a photon to the output file - this allows us to properly reconstruct the in cell spectrum. Again, photon noise changes to fiducial an run
5: Fixed bug in electron yield input - no change to spectra
6: Fixed error in calculation of schull type recommit data. Some changes to fiducial spectra, but within noise
7: Added some blocks in para_update to try and prevent iridis hangs. May still not be fixed! - no change to spectra
8: Also changes to hydro to allow restarts to use a past wind_save file to inform ionization state, added auger heating to mode 1 in py_wind, and added a mode 2 to output all ions for a given cell in ascii read-in style. Also a couple more bug fixes. None of these changed the AGN fiducial spectrum.
9: Added the correct library to python.c to allow the clock call in python to set a new random number seed.

### DIFF
--- a/source/Makefile
+++ b/source/Makefile
@@ -113,7 +113,7 @@ LDFLAGS= -L$(LIB) -lm -lcfitsio -lgsl -lgslcblas
 #Note that version should be a single string without spaces. 
 
 
-VERSION = 80
+VERSION = 80a
 
 CHOICE=1             // Compress plasma as much as possible
 # CHOICE=0           //  Keep relation between plasma and wind identical
--- a/source/atomic.h
+++ b/source/atomic.h
@@ -30,7 +30,7 @@
 #define ANGSTROM                        1.e-8	/*Definition of an Angstrom in units of this code, e.g. cm */
 
 #define EV2ERGS   			1.602192e-12
-#define RADIAN				57.29578
+#define RADIAN				57.29577951308232
 #define RYD2ERGS                        2.1798741e-11	/* Rydberg in units of ergs */
 
 
--- a/source/brem.c
+++ b/source/brem.c
@@ -35,7 +35,7 @@ integ_brem (freq)
      double freq;
 {
   double answer;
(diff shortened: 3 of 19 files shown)


jhmatthews:
@Higginbottom -- just to confirm -- this passed all your tests, right?

Higginbottom:
I ran the fiducial AGN model and my hydro model. Both agree to around the photon noise level.
CV model also runs, and I think its also within the noise...

jhmatthews:
Cool. I will run a separate test against py80 when I'm finishing my test suite anyway.

jhmatthews:
Nick @Higginbottom , I merged this. Note that I modified your random seed thing for Zeus slightly to

``` C
  /* initialize the random number generator */
  /* JM 1503 -- Sometimes it is useful to vary the random number seed. 
     Default is fixed, but will vary with different processor numbers */
  /* We don't want to run the same photons each cycle in zeus mode, so 
     everytime we are using zeus we also set to use the clock */
  if ( (modes.rand_seed_usetime == 1) || (modes.zeus_connect == 1) )
    {
      n=(unsigned int) clock()*(rank_global+1);
      srand(n); 
    }
  else 
    srand (1084515760+(13*rank_global));

```

See also #203

PR #207, opened 2016-03-31
Two modifications to hydro_2_python helper script

Higginbottom:
The first commit was a bug, the second allows the code to cope with increased precision in zeus time data.


jhmatthews:
Sorry Nick, missed this request! Merged.

Issue #170, opened 2015-08-24
Log spectrum


Higginbottom:
Trying to compute X-ray spectra over a wide range of frequencies, in order to see compton scattering has led me to think that we should be able to generate logarithmically binned spectra in the spectral cycle as well as the ionization cycle. This is a simple process, just requiring log bins to be implemented in extract.c

I also note that the linear and log tot_spec spectra are not comparable, since neither is scaled by the width of the frequency bins. I propose scaling these two spectra in this way, so the input spectra are specific luminosity.

I have already coded and tested this, I'll put it into the dev branch and close this issue after our Friday telecon - just in case anyone doesn't like it!

N

jhmatthews:
See also #149 - which I never got around to doing.

jhmatthews:
@Higginbottom - has this been coded?

jhmatthews:
This was closed by #213

PR #213, opened 2016-06-23
Implementation of log spectra

Higginbottom:
This pull request is for bug report #170.
log_spec is output as the logarithmic spectrum equivalent of spec
log_spec_wind is the log equivalent of spec_wind.

The spec_tot file is also slightly modified. It is now divided by dfreq so that the log_spec_tot file looks the same as the log_spec file. Both now show L_nu rather than L_nudnu


jhmatthews:
Looks good Nick :) Merging

Issue #180, opened 2015-10-14
Treatment of photons when hit disk


kslong:
Python, nominally, has 3 ways of treating a photon if it hits the disk.  It can be lost, It can be reflected, it can be absorbed and the temperature of the disk modified to account for the energy absorbed.  As far as I can, the last of these options is not working.  All options need to be checked.

kslong:
It looks as if geo.disk_illum is not set in  get_disk_parameters, which is part of setup.c.   There are comments that one cannot immediately set disk_illum because other items (e.g velocities in the KWD model have not been set yet.  get_disk_parmaters returns disk_illum, but python main does not "receive" or use them.  It is possible this got messed up when this routine was created in an attempt to simplify the main routine.

Note - while we do not recalculate the grid on restarts, we do reconstruct this disk and this is a restart issue if our approach has modified the disk temperature.  It seems likely that the disk structure should be saved.

Note that This whole routine is quite confused, with some bits commented out by Nick because he says they caused problems on restart.  But it is clear the never figured out what the underllying problem.  This is a good example of a band-aid, that is now harder to figure out than it would have been at the time.  Restarts, should not need to get material from the command line.

kslong:
These are the nominal meanings of disk_illum as given in python.h

  int disk_illum;               /*Treatment of effects of illumination on the disk. 
                                   0--> Disk simply absorbs the radiation and it is lost
                                   1--> Disk reradiates the radiation immediately via electron scattering
                                   2--> Disk radiation is absorbed and changes the temperature of the disk for
                                   future ionization cycles
                                   3--> Disk illumination is treated in terms of an analytic approximation
                                   04Aug ksl -- this parameter added for Python52

kslong:
This is now at least partially fixed.  In particular, geo.disk_illum should be properly passed to the downstream routines.  

Some issues remain however:
1.  For restarts, it the program will probably fail becasue the disk structures are not saved and read back
2.  One needs to check that the program works properly with multiple threads.  It is not obvious to me what is done with the structures for disk.

kslong:
Note - It is not obvious that we need a structure qdisk and disk.  By adding one more T to disk, which is the effective temperature after accounting for the radiation that has hit the disk we might be able to eliminate qdisk.  We have to be careful though with the way that the annuli of the disk are established or reestablished because this redefines the positions of the annuli.

kslong:
[...]

PR #217, opened 2016-07-12
Fix qdisk in parallel mode

jhmatthews:
This is a fix for issue #180 -- it means that the statistics for the photons that hit the disk are now communicated between the threads as they should be. This involves adding some extra loops and helper arrays in the function communicate_estimators_para () in para_update.c.

I've tested that it works. This plot shows the main quantities from the disk.diag file in serial mode, and then in parallel mode with and without this fix:
qdisk

The only quantity that changes in overall value is the number of photons that hit the disk -- the other quantities just get less noisy. This is exactly what should happen.

In the process, I also made a few other small changes:
- added a few MPI_Barriers in communicate_estimators_para() which should aid stability
- made the qdisk_save compatible with astropy.io.ascii.read()
- modify the free and calloc command order in communicate_estimators_para() to ensure we only allocate the minimum amount of memory needed at any one time (i.e. doubles, then integers, rather than all together).

### DIFF
--- a/py_progs/qdisk_plot.py
+++ b/py_progs/qdisk_plot.py
@@ -1,0 +1,62 @@
+#!/usr/bin/env python 
+'''
+    University of Southampton -- JM -- July 2016
+
+                qdisk_plot.py
+
+Synopsis:
+    make a plot of the quantities in the disk.diag file
+
+Usage:
+    Either import as a module in a python session e.g.
+    import qdisk_plot 
+
+    or run from the command line e.g. 
+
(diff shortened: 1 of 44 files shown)


jhmatthews:
I've also now added a plotting routine to py_progs which allows you to easily plot the above quantities by running

```
qdisk_plot.py root
```

if you have a file named diag_root/root.disk.diag in your directory.

jhmatthews:
Tests all look good, merging 👍

PR #231, opened 2017-01-31
Fix how we deal with extra diagnostics

jhmatthews:
I've rewritten get_extra_diagnostics and added some functions to rdpar. The basic idea of this is the following.

If you want to operate extra diagnostics - for example, getting in cell photon statistics or adjusting the grid scale lengths - then you need to run python with the following command for the cv_standard model

```
py -d cv_standard
```

You then specify the extra diagnostics in a file called ```cv_standard.extra``` which python looks for on start up. If it isn't there, it doesn't worry about extra diagnostics. If it is, it reads it into a structure just like the normal input file

Then, each time it gets to a point in the code where it needs to ask an advanced question, it looks in this file for the relevant entry. It doesn't care whether things are in a certain order but does check for duplication.  

**To test / use:**

I've done this in a new branch called extra_diags. Checkout that branch and try out the code - the example files in $PYTHON/examples/core/ should be up to date with the code. You can do this with something like:

```
git fetch upstream
# This will fetch all of the remote branches for you. 
# You can see the branches available for checkout with:
git branch -v -a

# With the remote branches in hand, you now need to check out the branch you are interested in, 
# giving you a local working copy:

git checkout -b extra_diags upstream/extra_diags
```


jhmatthews:
Main issue here is to deal with the following options in the code, from setup2.c.

```C
  if (modes.iadvanced)
  {
    strcpy (yesno, "n");
    rdstr ("Select_specific_no_of_scatters_in_spectra(y/n)", yesno);
    if (yesno[0] == 'y')
    {
      Log ("OK n>MAXSCAT->all; 0<=n<MAXSCAT -> n scatters; n<0 -> >= |n| scatters\n");
      for (n = 0; n < geo.nangles; n++)
      {
        rdint ("Select_scatters", &geo.scat_select[n]);
      }
    }
    strcpy (yesno, "n");
    rdstr ("Select_photons_by_position(y/n)", yesno);
    if (yesno[0] == 'y')
          Log ("OK 0->all; -1 -> below; 1 -> above the disk, 2 -> specific location in wind\n");
      for (n = 0; n < geo.nangles; n++)
      {
        rdint ("Select_location", &geo.top_bot_select[n]);
        if (geo.top_bot_select[n] == 2)
        {
          Log ("Warning: Make sure that position will be in wind, or no joy will be obtained\n");
          rddoub ("rho(cm)", &geo.rho_select[n]);
          rddoub ("z(cm)", &geo.z_select[n]);
          rddoub ("azimuth(deg)", &geo.az_select[n]);
          rddoub ("r(cm)", &geo.r_select[n]);

        }
      }
    }
  }
```

The problem is that I've not required the ```.extra``` file to have all the entries like in the parameter file, and so I can't force it to have a certain number of options.

Proposed solutions:
* make the  ```.extra``` file have a fixed format and prompt the used if I can't find any specific variables
* write additional routines to read below the options if they require additional arguments.

Note that a lot of these problems are caused because the way rdpar is written doesn't really support reading from a different file / structure. One could provide the structure to read from as an argument, or have a mode setting, because at the moment it is not general. 

Any thoughts @kslong @smangham @Higginbottom  on ease of use?

kslong:
Even though I know we have discussed this in the past, I really wished we had rediscussed it again before implementation.  I had thought that the fact that one can now read inputs in any ordered obviated the need for a separate file.  I presume that if the .extras file is missing, one does not actually require it to get the “extra” diagnostics, rather that both the normal file and this are read in and then treated as if it were a single file.  I certainly opposed to a situation where one has to choose what goes in the extras file and what goes in the main .pf file.

As an aside, I believe that if you are going to have this, it should still be a file with the extension .pf … so that we don’t promulgate additional file types.

jhmatthews:
Although i probably favour having a separate file, perhaps the best thing to do is to just have dollar signs, or similar, in front of advanced commands as you have suggested before Knox, and put them at the end (but allowed anywhere) of the file. 

Note that the major change in the code isn't only having an extra file - but also allowing the situation where advanced commands don't all have to be in the file when you run with advanced - so some of the above questions still apply, namely how we deal with *optional commands* that have *required follow-on questions* if they are true. 

If you agree, I can code this up.

jhmatthews:
JM, KSL and NSH decided in a telecon to just adapt the current mode of operation rather than deal with separate files, so that the "prompting" for missing answers capability is retained. This pull request can thus be closed without merging.

Issue #250, opened 2017-03-28
Segfault in YSO model


jhmatthews:
This is an issue being reported by Kelly Milliner, an undergrad student of Lee Hartmann

**Environment**
Mac, no mpicc available so running serial.

```
gcc (MacPorts gcc49 4.9.3_0) 4.9.3
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

```
Python Version 81b_dev
Built from git commit hash f6111a26e393788db2a9ca4e2224c733a7341f5a
```

**Input file:** yso_a.pf from this repository: https://github.com/jhmatthews/fu_ori
**Output with segfault:** [output.txt](https://github.com/agnwinds/python/files/876463/output.txt)

If anyone is able to run with these exact settings or has any ideas I'd appreciate it - I can't reproduce the error, which I've tested with the below compilers I have available to me, and it's holding up Kelly's undergrad project a lot.

@smangham @Higginbottom @kslong 

Tested with [all using the above commit of python]:
* On mac with Apple LLVM version 7.3.0 (clang-703.0.31)
* On mac with gcc-mp-5 (MacPorts gcc5 5.4.0_1) 5.4.0
* On linux with gcc-5.4.0
* On linux with gcc-4.9.3

kslong:
This does NOT fail for me on my MAC  either using

gcc version 5.4.0 (GCC)

As it exists on Fink

Or

Apple LLVM version 7.3.0 (clang-703.0.31)

Knox

Higginbottom:
I can’t get the fault any more - 
now compiling using my version of GCC on the mac which is 

Apple LLVM version 7.0.2 (clang-700.1.81)
Target: x86_64-apple-darwin14.5.0
Thread model: posix

jhmatthews:
so you can't reproduce it at all anymore? This is really weird. I haven't been able to reproduce it with any of the compilers I've tried (now over 10). If anyone can help it'd be good because we aren't making the best first impression at the moment. Nick, do you know if you get it when you do the same steps with dev latest?

kslong:
Have we checked Nick’s suggestion, namely that the student pull the current version, do make clean, recompile, and try it again. We definitely need to be able to reproduce the error to fix it, but I’d prefer not to have to figure out how to install a macports version of gcc.  Knox

[...]

PR #251, opened 2017-03-30
Collision strength plus bug fix for #250

Higginbottom:
Two main things in this update
Collision strength data from Chianti read in in read_atomic data and applied in lines.c
also fix to #250 simple protection in wind_updates2d.c so if there is no change to t_r or t_e, then the code reports this rather than attempting to call wind_n_to_ij which was the cause of the segfault


jhmatthews:
This looks good Nick. Do we know how it affects various models / tests?

Higginbottom:
not yet, 
I’ve done IP loops on the flat spectrum from Mehdipour and it is doing what I want it to do. Just about to run some models. It would be great to spend a bit of time chatting tomorrow about the test runs. I think getting this suite working is top priority!!!

Higginbottom:
discovered that my new variable length plasma arrays were not correctly implemented in the MPI passing in wind_updates2d.c Now fixed

PR #252, opened 2017-04-25
Fix advanced mode

jhmatthews:
This initial fix to the currently broken extra diagnostics is pretty simple. Remove the slightly convoluted way of reading advanced diagnostics I had written which meant the functions didn't have to be there, and replace with rdint() functions for each option.

To improve it slightly, I suggest we adopt Knox’s idea of prefacing each advanced command with a special character.

This also makes it quite easy to add extra diagnostics - you just need to add the relevant code logic, an option in the "modes" structure, and a question in get_extra_diagnostics(). It's possible that 
some of the current normal modes should be moved here (e.g. standard_care_factors).

Thoughts welcome.


kslong:
For a special character, I suggest @.  Note that #, !, $ are all taken.  Knox

PR #263, opened 2017-06-21
Upstream low temp

Higginbottom:


Higginbottom:
Initial work to incorporate a properly calculated radiative recombination luminosity. Name changes to ensure that only variables that are used to generate photons are called lum_something, otherwise they are called cool_something.

Issue #265, opened 2017-06-23
Correctly reporting absorbed flux


Higginbottom:
One of the useful diagnostics that is reported as python is running are the Absorbed flux, Wind heating, Wind luminosity and Wind cooling lines. 
The Absorbed flux is incorrect, since it actually only reports the wind cooling for photoionization. This is particularly important in testing the new recomb cooling/lum fix. 
I'm going to try and implement a more useful value which will be the photon flux absorbed by PI processes - i.e. including the binding energy.

jhmatthews:
Good idea 👍

Higginbottom:
fixed by pull request #266 

Very interesting.

PR #266, opened 2017-06-23
Added abs_photo, abs_auger and abs_total fields to plasma structure. …

Higginbottom:
…These store the photon energy absorbed, as opposed to the heating. This allows these things to be reported in an attempt to understand why we are seeing more energy coming out of a model with the thick wind than we put in


Higginbottom:
This addresses bug #265

PR #278, opened 2017-07-26
Re-commit of the changes to CDF calculations.

Higginbottom:


Higginbottom:
This is a re-commit of the changes to CDF calculations. 
It mainly affects cdf_gen_from_array, but there are many minor changes in other routines. All the comments are not yet fixed up.
There are issues rating to blackbody CDF generation which I noted when doing careful testing, these are listed in #281 #280 #279
Other bits are some changes to the BB CDF - there were forced steps in the CDF which were outside the range where the CDF was generated - almost certainly my fault. Also there is a change to emittance_bb (in bb.c). There was an obvious bug here which was calling the tabulated BB function even when part of the band was outside its range.

Issue #284, opened 2017-08-01
Mismatch between _cool, and _lum plasma fields 


Higginbottom:
A bit of an old chestnut this one, but re-surfaced due to the new split in rr_cool and rr_lum. 
There are a few different fields in the plasma structure that deal with cooling and luminosity and these are a little confused. More confusing is when they are updated prior to being written out into the wind_save file. 
This problem specifically crops up with the way in which there are now two radiative recombination fields, lum_rr (the radiative recombination luminosity) and cool_rr (the RR cooling). The former is updated at the start of a cycle, when photons are made (lum_lines, lum_ff and lum_tot are also filled at this time) and ends up containing the band limited luminosity for the last band generated. This is frequently a rather high frequency band and there is no luminosity, so probably all four fields end up containing a zero.
The photons are flown, and then the effect of those photons is computed via one shot. At this point, lum_lines, lum_ff, and *cool*_rr and *cool*_tot are populated during the search for the new temperature. NOTE - these fields end up populated with the rates at the last temperature used in the *search* for t_e, not necessarily the actual new temperature.

After this, the processes all exchange information, and then the _ioniz fields are populated.

Then wind_cooling and wind_luminosity are called to update the cooling and luminosities to the current t_e, and  then the wind is reported and exported.

So, why is this confusing?

1: The _ioniz fields don't contain the same information as the fields they are supposed to be copies of if the temperature has changed

This caused me a lot of headaches, since I was exporting a .complete file from py_wind, then picking a cell to examine more closely using the 'e' option in py_wind. The former uses the 'normal' fields, e.g.. lum_rr, whilst the latter uses the _ioniz fields - e.g. lum_rr_ioniz. The fact that these were different was very confusing!!

Solution?

Call wind_cooling in one_shot *after* the new t_e has been calculated to ensure that right from the outset the _cool fields have the cooling rates that match the actual adopted t_e rather than a random t_e in the search. This would also mean that the convergence criteria is computed for the new t_e...

BUT - we need to be more careful about the _lum fields

we should decide if we want the _lum fields to represent the luminosity of the wind in its new state, or in the state where the photons were generated. I think we probably want it to represent its current state - otherwise it will get very confusing. So this is fine, but we need to move the call to wind_luminosity in wind_update to before the point at which everything gets copied into the _ioniz fields.

its also a little confusing that we use lum_lines for cooling and luminosity - its kinda fine since they will always contain the same value though.

kslong:
[...]

Issue #270, opened 2017-07-07
python no longer working with stellar models


kslong:
On the current dev branch, python is unable to generate cdfs for either a star or a disk when Kurucz models are used instead of bb.  

@lazygun37 , @Higginbottom is correct that python is not working with stellar models for a star.  

These errors are among those being generated:

   100001 -- pdf_check: x problem n %d x %f pdf->x[n] %f
        7 -- error_count: This error will no longer be logged: %s
      502 -- pdf_check: x values in pdf should be monotonic
      502 -- pdf_check %d %d %d %d %d
      502 -- pdf_gen_from_array: error %d on pdf_check
      502 -- In one_continuum after return from pdf_gen_from_array
      503 -- pdf_get_rand: %d
      503 -- pdf_get_rand: %g %d %g %g

Higginbottom:
This is more complex hat it first seems...

Here is a file which fails:
[stellar_1em16_kuruz.pf.txt](https://github.com/agnwinds/python/files/1173126/stellar_1em16_kuruz.pf.txt)

however this one - which uses the same kurukz file runs fine
[star.pf.txt](https://github.com/agnwinds/python/files/1173129/star.pf.txt)

There are many differences between the files, and its possible that the first one breaks all kinds of rules about what we should do with models, but it should fail gracefully..

kslong:
@Higginbottom has "settled" more or less on a new formulation of levels in cdf.p, but as of today, the following input files sv.pf file that ksl first tested is still failing with the same error as above.

[sv.pf.txt](https://github.com/agnwinds/python/files/1185933/sv.pf.txt)


Nick reported that he has some files that do seem to work with kurucz models and some that do not.

star.pf above has no ionization cycles, but there are no problems even if these are added. There are also not problems seen when the T is lowered to 1000 K or raised to 100,000 K which is outside the grid.  


The stellar_1em16_kurucz.pf file does fails, but if fails for different reasons than sv.pf does, viz

one_continuum: f too large %e

... which is weird.  Somehow what is happening is associated with using kurucz models though because if you drop back to a bb spectra, you do not get errors with same file.

Note however, a little debugging revealed that the code is going through one_continuum in both cases, so the errors are not completely unrelated.

**When exactly did the problem appear:**

It was present

commit 20682080a2ce4c07e1106bfa71e17fd94525bf97
Merge: 872490e 365f4d5
Author: jhmatthews <jimim@hotmail.co.uk>
[...]

Issue #261, opened 2017-06-20
Cooling v Luminosity revamp, including bug in bound-free photon generation


jhmatthews:
We have discovered a major issue with bound-free photon generation, in that we currently use lum_fb as our "bound-free wind luminosity" -- i.e., actual radiative energy emitted by the wind, but lum_fb is calculated from total_fb which gives a "bound-free cooling rate" -- i.e. energy transferred from the thermal pool. 

@kslong , myself and @Higginbottom are going to work to improve the code both in terms of clarity and commenting, but also we are generally going to distinguish explicitly between cooling rates and radiative luminosities. As a result, new variables are required in the geo and plasma structures, and we will probably n

The actual bug in bf photons will require that lum_fb gets the integrated bound-free emissivity from

A few examples:
* lum_adiabatic should really be cool_adiabatic
* lum_fb should give the radiative luminosity, cool_fb should give the cooling rate
* lum_ff and cool_ff would be the same. It is not clear yet if we should have both variables in this case.

This work is being carried out in the low_temp branch. Nick is going to start working on the process of converting variables. I will be responsible for making sure everything works with macro-atoms. We should try and document here.  

This is pretty much priority A for me Knox and Nick at the moment as it needs to be fixed before new meaningful science can be done.

kslong:
@Higginbottom has now updated the structure names, though since he has done this via sed  a lot of comment updates are needed.  Knox has created a new set of routines for cooling in a new file called cooling.c and updated the total_emission to call fb_total in FB_FULL mode.  The revised version of the program on the low_temp branch runs but produces identical results to what was produced before any of the structure updates.

Knox believes the problem is likely associated with the fact that we store results in a set of fb arrays and if the frequency range of interest is the same as the last time fb_total was called, then it uses the values stored in the arrays instead of calculating them with a new mode.  However, this needs to be verified.  More generally, we need to think how these arrays are created and used.  The original justification for storing the arrays was that calculating them was thought to be time consuming, but I suspect that may not be a good justification any long.  In any event, some thought should go into this by @jhmatthews and the rest of us.

Higginbottom:
Storing the arrays is still definitely needed. When I implemented my inner shell rate calculations (to compute dielectronic recombination) the code ran exceptionally slowly before I included a new array for inner shell recombinations. I believe that the code works are required, and I can't see an obviously better way of doing this.
[...]

Issue #258, opened 2017-06-05
How luminosities are handled


kslong:
I have been trying to construct lum_tot to verify that we are actually getting the right answers for our ionization balance.  What I find however  looks like a mess in the code, where total_emission does not include all of the cooling processes, etc.  I'd like comments from @Higginbottom and @jhmatthews before I do much because I want to make sure that we are consistent with one another.  

When I say things are a mess, what I mean are lines in things like zero_emit, where we seem to compute the heating balance

```
 difference = xxxplasma->heat_tot - xxxplasma->lum_adiabatic - xxxplasma->lum_dr - xxxplasma->lum_di - xxxplasma->lum_comp - total_emission (&wmain[xxxplasma->nwind], 0., VERY_BIG);
```
Why is it that total emission does not include all of the processes, especially lum_comp.  Do we need a new variable cool_tot to capture the totol cooling.  Is lum_rad supposed to capture the total_emission?

The confusion extends to python.h where we have variables like 

```
 double lum_dr;                /* 1109 NSH The dielectronic recombination luminosity of the cell */
```
Is this a luminosity or a cooling term.  

And another question for each you.  Is zero_emit always called when we are trying to balance heating and cooling, and so is this the place we should actually be calculating all of these things and storing them in the plasma array so we can see the state at the end of an ionization cycle?

More generally, can you all suggest how we might make all of this clearer in the code, as misunderstandings here will certainly affect how easily if will be to sort out what is going on in the optically thick case (where different processes make relatively more contributions than we are typically used to).

jhmatthews:
I think the main point here is that lum_dr, lum_comp etc are not used to actually produce wind photons, so are just cooling terms in this context. I believe lum_comp is essentially taken care of in the actual scattering event, by adjusting the weight and frequency of photons that "upscatter" in the plasma. I am not sure what we currently do with DR as I thought we now produced DR photons via Milne+Auger cross-sections. DI is a way for thermal energy to be converted to ionization energy and so doesn't create photons directly. I agree the whole thing is a bit of a mess, at least semantically if nothing else.

[...]

PR #293, opened 2017-09-04
Low temp2

Higginbottom:
This is a major pull request, to pull in the work done by NSH and KSL in the summer of 2017 to address problems with python working in optically thick conditions. 
The following issues are addressed

#261 - This was a major bug - free-bound luminosity did not include the ionisation potential of the recombining ion, so the luminosity was far too low and energy was being lost. This caused cells to cool to very low temperatures. A set of new luminosity variables was made to allow radiative recombination to be split into a cooling and a luminosity variable. New code made the 'full' mode in the various subroutines contained in recombination.c work as intended. Many other commits fixed the various checks for convergence, sharing between parallel tasks and outputting the data in wind_updates2d.c. Py_wind was also modified to output these variables as part of the 'e' option and the '1' option.

#258 - this is really a question, but is addressed by the changes made to split luminosity and cooling

#270 - This was fixed during the major revamp of PDFs 

#284 #36 These two issues raised about when we calculate the cooling rates which will be saved into the wind save file have been addressed. We now compute the cooling rates one last time after the new electron temperature has been settled upon.

Other important things dealt with that were never raised as issues:

The initial reason for this new branch was KSLs experiment with reducing the minimum temperature at which FB cooling was calulated. We now compute FB cooling down to 100K

Other changes were made by KSL at the start of this work that he might be best placed to explain - talking about the modifications made up to about Jun 20...

The other major piece of work was prompted by testing of the new recombination code which revealed very odd shapes to recombination continua in a h only model. This was tracked down to problems in the code for making recombination photons. The original code was intended to use an array of 'jumps' which told the CDF code where discontinuities were in the freebound continuum. This was broken and has essentially been bypassed years ago. It was fixed by eliminating the jumps and instead ensuring that the PDF was properly sampled around discontinuities. This necessitated a significant rewrite of the PDF-CDF code and this opportunity was used to also rename many subroutines to properly indicate wether they were dealing with PDFs (which most weren't) or CDFs (which most were - so many routines were renamed from PDF***** to CDF*****. Changes were also made to free free generation and models - this is where #270 was fixed.

Other changes were made to deal with the extra variables that were made - writing out in py_wind and communicating for MPI.


jhmatthews:
@kslong - most recent change seems to produce a segfault. Should one use fclose(file_ptr) if file_ptr is NULL? I think that's the cause.

kslong:
Mea culpa.  Should be fixed now.

jhmatthews:
I've attempted to check that this looks sensible, but it's not an easy task. In particular, I have not checked changes to cdf.c, but all the other changes look like they are at least based around the right problems. Testing to continue, but merging into dev now.

PR #312, opened 2017-11-08
Corrected a small bug in the velocity interpolation for zeus import

Higginbottom:
When investigating a few things in the hydro import module, I noted that the interpolation in velocity was not quite correct. v_r in the hydro model is defined at the theta edge of the cell, and the r centre, and vice versa for v_theta. v_phi is set at r and theta centres. Highly unlikely to make the slightest difference to anything, but important to be correct!

### DIFF
--- a/source/hydro_import.c
+++ b/source/hydro_import.c
@@ -370,34 +370,51 @@ hydro_velocity (x, v)
     theta = asin (xxx);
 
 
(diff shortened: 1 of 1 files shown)


jhmatthews:
I guess we want this merged @Higginbottom ?

Higginbottom:
I thought I had :-)

jhmatthews:
Just checking. merged!

PR #328, opened 2018-01-31
Fixed error in attempt to debug changes made to cdf,c

Higginbottom:


Higginbottom:
This is a change made to debug a change made to mitigate some of the changes KSL made previously. There was a mismatch between the array limits generated by KSLs new tests and the CDF generation code later in the subroutine.

kslong:
This is not detailed enough to comment on.  Please provide details of what was the problem previously.  @Higginbottom saying it was to "mitigate some changes" does not make clear what the problem you were trying to address was.

Higginbottom:
This check
  nmax=nmin;
  while (x[nmax]<xmax && nmax<n_xy) {
      nmax++;
  }
will leave nmax=n_xy if (as is usually the case) the input array conforms to the required array, i.e. x[0]=xmin, and x[n_xy-1=xmax.

This causes problems later on with

cdf->y[cdf_n]=cdf->y[n-1]+y[nmin+cdf_n]*(x[nmax]-x[nmax-1]);  //This interpolates to get the last point in the CDF

so the extra check dealt with this case. The lack of a ; meant that the fix of reducing nmax by one was always applied, rather than only in the case where x[nmax-1] was actually less than Xmax. I think this test will always be true anyway...

Issue #239, opened 2017-02-09
Apparently different random numbers on different machines


Higginbottom:
I've noticed subtle differences in simulations carried out on different machines. In particular heating rates computed directly from photon fields. In partcular:
My Macbook = My iMac
 Endjinn = Iridis
My macs != the Iridis.

Checking I note that the two classes of machine generate different photons. 

If this is is just down to different random number lists then this is not necessarily a problem, but should be noted. 
If it is however down to something different in the code, then this is a worry.

Higginbottom:
I've tested a simple random number code on my mac and endjinn.

There are two possible random number generators rand() and random()
I have two compilers on my mac gcc (4.8.4) and clang (7.0.2)
One compiler on enjin gcc (5.4.0)

both commands give the same set of random numbers on endjinn

the two compilers give the same results for each of the commands on the mac. But the two commands give different numbers, and neither set are the same as on endjinn.

Hmmmmm

OK, so I then got a 'platform independent' random number generator - the Mersenne Twister - and coded that up. It gives the same set of random numbers for all compilers and all machines.

On the web, people are not confident that rand() should give the same results cross-platform. So, the question is, do we implement a MT style random number generator in python to assure identical photons for the same seed on different machines. I should really test to se that with a MT random number generator I get identical results on various machines, to ensure that the differences I saw originally are indeed explained by the random number 'issue"

smangham:
So whilst gsl has a Mersenne Twister generator it sounds like the SFMT implementation is better- twice as fast, and not as vulnerable to being poorly configured: http://www.math.sci.hiroshima-u.ac.jp/~m-mat/MT/SFMT/speed.html
There's libraries here, including ones aimed specifically at generating doubles that are faster. In particular, it has SIMD versions which seem to use *'Single Instruction, Multiple Data'* capabilities added to CPUs in the early 2000s. It's probably safe for us to assume that most computers we're using can do them, but the MT implementation in GSL seems to be from ~2002 so probably does not use them.
http://www.math.sci.hiroshima-u.ac.jp/~m-mat/MT/SFMT/
Given how little time random number generation takes a ~4x speedup may not be worth the effort of including a new library, though.

Higginbottom:
Here is my tiny c program, and some data files generated by doing

gcc --version > file.dat
rand_test >> file.dat

I had to change all the filenames to txt to allow git to upload them......


[rand_test.txt](https://github.com/agnwinds/python/files/767474/rand_test.txt)

[...]

PR #341, opened 2018-02-09
Random

Higginbottom:
A pull to fix issue #239. The intention is to make the sequence of random numbers the same for all platforms

I've been quite conservative, and replaced rand() /MAXRAND with something that looks like gsl_rng_get(rng)/randmax where
randmax is the maximum random number from the generator obtained from randmax=gsl_rng_max (rng).

rng is a pointer to the random number generator which is set up with rng = gsl_rng_alloc(gsl_rng_mt19937) in python.c

rng is made global by putting in in python.h

I have left in all the lines I've tried to replace, commenting them out and adding a DONE to the end of the line. This will be fixed soon, once I'm sure all is working correctly. I've also updated the version to 82f


kslong:
This seems like a fairly major change to me, as it affects many parts of the program.  Nick, you have been very reluctant to make changes until you are finished with the X-ray binary paper.  Is this really what you want to do.

Knox

Higginbottom:
That’s why I’ve changed the version number. I’m running py82d for the Zeus Python. 82e is the new random number. So getting the random number working is not holding up the paper. Indeed if I can’t work out why the travis build isn’t working I’ll give up!! 
N

PR #363, opened 2018-04-03
commented dielectronic.c

Higginbottom:


Higginbottom:
doxygened and commented dielectronic.c

PR #404, opened 2018-08-02
Multiple cleanups: rm CHOICE, rm SCATTER_MODE_ANISOTROPIC, streamline line transfer modes

jhmatthews:
This pull request addresses a few XXX comments and outstanding issues.
* It removes SCATTER_MODE_ANISOTROPIC, an approximate anisotropic scattering mode that we never used. This involves the removal of a few functions in anisowind.c which also contained some other XXX comments
* It stops the dangerous practice of reading in geo.line_mode from the command line and then modifying it by if statements, instead having a local variable user_line_mode.
* It removes CHOICE from the Makefile and any routines which use it (we always set it to 1 anyway).
* It also allows one to run in mode 9, which forces the use of all simple ions but with SCATTER_MODE_THERMAL
* it prints a help message to the user if they request a deprecated or incorrect mode 

**Testing**
In order to test this, I ran $PYTHON/examples/basic/star.pf 9 times with the 9 different modes. All the modes produced identical output files except the deprecated mode 4 and the slightly modified mode 9. This is the expected behaviour.  

Given as this is a fairly major change, it would be good for someone to review it: @Higginbottom @kslong


kslong:
I have also removed CHOICE, etc..  My version is merged into dev, which is why there are conflicts I suspect.  The changes look identical, except that I may have had slightly more discussion.

kslong:
I believe I have now merged this properly into dev.  It took two tries.

PR #437, opened 2018-09-24
Use comoving frequency when calculating bound-free kpkt probability for simple atoms 

jhmatthews:
Would be good if someone could check this looks sensible; I haven't tested this at all yet @kslong @Higginbottom

### DIFF
--- a/source/matom.c
+++ b/source/matom.c
@@ -501,7 +501,6 @@ matom (p, nres, escape)
 
       
       /* Co-moving frequency - changed to rest frequency by doppler */
-      /*Currently this assumed hydrogenic shape cross-section - Improve */
     }
     else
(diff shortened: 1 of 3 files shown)


kslong:
@jhmatthews   Thre is at least one error in this that is causing a segfault with short_34.pf.  It is in resonate.  (This is in a different place than the "normal segfault" in mp mode)

screenshot_840

You have not defined ndom in line 1011.

jhmatthews:
Fixed that - did a force push to update (naughty).

kslong:
@jhmatthews   I think you could merge this now.  So far as I can tell this has "solved" that problem in the sense that errors are being trapped and the program continues onward.

Issue #425, opened 2018-08-19
Carbon macro atom models are failing


kslong:
This is an issue that Keara discovered.  While the CV models seem to be working with her C34 and C345 macro atom models, the AGN models do not.  Her discussion is  [here](https://www.evernote.com/l/AS-LP8srzLpDq7zU0Rj32pjlO4Ndb5b-XnU)

ksl confirms that some of these models are failing.  In particularly, running in multiprocessor mode, fiducial_agn_c34.pf had a segmentation fault.   The same model running in single processor mode, died after too many sane_checks, as did fiducial_agn_345.pf.  It would seem as if the first step is track down the cause of all of the sane checks.

kslong:
The model.pf files that caused this problem have to be updated to reflected the fact that pairwise_pow has been reduced as an ionization mode. 

After I made these modifications and I still see problems in multprocessor mode, but it seems to be running fine in single processor modes.  The failure mode was not quite the same for the 34 model ther was a segmentation adult in multiprocessor mode, for the 345 model I got a more generic mpi failure message.

kslong:
I tried running in multiprocessor mode with a smaller wind (30 x 30) instead of 100.  This time in mulitprocesor mode I still get a segemetatin fault for the 34 model.  The last lines of the output file before the segfault look like this:

```
Thread 0 happy after broadcast.
!!python: Number of ionizing photons 2.03478e+56 lum of ionizing photons 6.34341e+45
mc_estimator_normalise: bb stimulated correction factor is out of bound. Abort.
stimfac -nan, i 26, line[config[i].bbu_jump[j]].nconfigu 29
estimators: den_config (xplasma, i) 0  den_config (xplasma, line[config[i].bbu_jump[j]].nconfigu) 0 
mc_estimator_normalise: bb stimulated correction factor is out of bound. Abort.
stimfac -nan, i 27, line[config[i].bbu_jump[j]].nconfigu 30
estimators: den_config (xplasma, i) 0  den_config (xplasma, line[config[i].bbu_jump[j]].nconfigu) 0 
mc_estimator_normalise: bb stimulated correction factor is out of bound. Abort.
stimfac -nan, i 29, line[config[i].bbu_jump[j]].nconfigu 30
estimators: den_config (xplasma, i) 0  den_config (xplasma, line[config[i].bbu_jump[j]].nconfigu) 0 
init_freebound: Creating recombination coefficients
[...]

PR #438, opened 2018-10-04
Minor fix for macro atom pop solver

ssim:
Possible solution to current issue on #425

### DIFF
--- a/source/macro_gov.c
+++ b/source/macro_gov.c
@@ -649,10 +649,21 @@ macro_pops (xplasma, xne)
 
             xplasma->density[index_ion] = this_ion_density * ele[index_element].abun * xplasma->rho * rho2nh;
 
+            /* JM Nov 18 -- to maintain consistency with the higher level routines, 
+               only allow density to drop to DENSITY_MIN */
+            if (xplasma->density[index_ion] < DENSITY_MIN)
+            {
+              xplasma->density[index_ion] = DENSITY_MIN;
+            }
+
             /* Check the sanity and positivity of the level populations */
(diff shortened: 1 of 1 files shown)


jhmatthews:
Hmm. Curiouser and curiouser. So this fix appears to have introduced an error into the c34.pf model that wasn't there before. That's a bit of a mystery to me...I'm on my way to visit @ssim so we will have a chat about this.

jhmatthews:
@kslong - fix made Knox, sorry for delay, we now catch the situation that ion density is zero. Could you try the test run out?

kslong:
I’ve begun the tests.  I will let you know how they turn out.

kslong:
Sigh @jhmatthews  ... this time fiducial_agn_c34 ran to completion but fiducial_agn_c345 did not (with mpirun -np 8).  It failed during Cycle 13 in thread 4 with this error.

Error: matom: macro atom level has no way out 11 0 0

There were no log/error statements before this error to indicate why it failed.  It was midway through transferring photons through the wind.

jhmatthews:
Thanks Knox - I think this is the problem I was seeing before I got ran into the restart issue (and I wasn't sure if it was to do with reading in the atomic atomic data). It means that somehow a macro-atom is being activated into it's true ground state i.e. energy = 0, bottom state of CIII, which should never happen because an activated state should always be the upper level. The question is then how it manages to get into this level in the first place.

Since all jumping probabilities are proportional to lower level energy (see Lucy or Stuart's talk from Belfast), once it gets to this incorrect activated state it will produce the above error.

jhmatthews:
Any ideas how that could happen @ssim ?

jhmatthews:
I think this pull request should be merged in any case, and discussion can continue on #425 - agree @kslong ?

Issue #439, opened 2018-10-05
Exiting due to errors


saultyevil:
Whenever we exit from Python due to an error, we generally exit in the standard C way using the `exit` function, with the error code 0. Usually, exiting with 0 indicates that the program has exited after running successfully with no errors. But in Python, exiting with 0 means both a successful and unsuccessful exit. This is slightly confusing to users and very confusing to `mpirun`.

`mpirun` will kill all processes if it notices that a process has exited with a non-zero exit status. However, as errors in Python exit with 0, `mpirun` doesn't detect that a process has exited due to an error and assumes everything is fine. This eventually leads to communication problems, such as causing Python to deadlock. It can also happen that Python appears to exit correctly, but MPI processes still continue to run in the background hogging resources; which is bad manners on a shared machine.

A fix for this problem would be to use a wrapper function to exit from Python depending on if the code has been compiled with `-D MPI_ON` or not. The most basic approach could look something like this,

```c
/*
 * Note: error_code should be positive and non-zero.
 */

void Exit (int error_code)
{
  #ifdef MPI_ON
    Log ("MPI Exit: MPI is exiting all processes with error code %i\n",
         error_code);
    MPI_Abort (MPI_COMM_WORLD, error_code);
  #else
    Log ("Exit: Python is exiting with error code %i\n", error_code);
    exit (error_code);
  #endif
}
```

But, it seems sensible to output more information than just aborting all the MPI processes. For example, JM and KSL proposed that Python should first flush the log and then output an error report to the logfile and exiting MPI more gracefully with `MPI_Finailze`. 

This should have the benefit of making error checking simpler and more robust, as well as avoiding any confusing exits in multiprocessor mode.

kslong:
@saultyevil  There are problems with your exit function when python is invoked with command line options that are supposed to to print out a little information and stop, e.g

py -h
py --version

and I suspect when one decides to abort in a interactive read in of the parameters.

Because we have gone through Exit these produce extra spurious outputs that we do not want.  In these case, we need to be able to silently close things off, not print out any error summary, etc.  It's possible we should use Exit(0) for this kind of close, and Exit(non-zero) for all of the others.  At any rate I think you have a bit more work to do on this.  Sorry!

saultyevil:
[...]

PR #449, opened 2018-11-26
Wrapper function for exiting

saultyevil:
Created a wrapper function to exit from MPI with more safety as outlined in Issue #439. I have also *hopefully* updated all of the previous calls to the standard C exit function to the new wrapper function.


saultyevil:
Ahh, the Travis build failed because I had a rouge `#include <mpi.h>` and Travis builds with gcc. I've fixed the issue in my fork and Travis seems to checking again.

PR #450, opened 2018-11-27
catch negative densities and false pop inversions from density floors

jhmatthews:
This change attempts to take care of situations where a density could be zero, 

The way it does this is use if statements involving DENSITY_PHOT_MIN (1e-10) to ensure that we only bother computing stimulated correction factors if the density of both levels is above this. It's possible this threshold is too high and should be DENSITY_MIN instead. 

@kslong - could you look at doing something similar with the Sobolev optical depth and perhaps push to this branch (macropops-fix), and also review my change, perhaps along with @ssim ?

### DIFF
--- a/source/estimators.c
+++ b/source/estimators.c
@@ -432,7 +432,7 @@ mc_estimator_normalise (n)
   double volume;
   int i, j;
   double stimfac, line_freq, stat_weight_ratio;
-  double heat_contribution;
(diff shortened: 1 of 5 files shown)


kslong:
I have made a change a pushed it back to macro-pops-fix.  I’m currently running the test cases.

PR #478, opened 2019-02-27
Improvements to kpkt photon generation

jhmatthews:
As discussed recently, we want k-packet photon generation to be non-banded and take place after the banded photon generation. This change implements this and has a new variable for the fraction of photons to reserve for k-packets. 

@kslong - I have run some initial tests of this with the pf file on "FU Ori - 2018 - Shock heating with k-packets" - the results seem sensitive to banding and to the number of k-packet photons and it's not clear if it's really working yet. The model is a strange situation since most of the luminosity above 13.6eV comes from the shock heating, since the star is so cool. 

I am not sure if convergence is being done the right way. I think heat_tot does not include heat_shock and so should have it added on, so I've done this, but I'm confused because I thought we already sorted this out. 

I am also unclear as to exactly how we use the photon numbers in photon generation, even in dev. We have an nphot_tot variable that is passed to define_phot, but also an NPHOT variable. I had to change these so they had the kpacket photons subtracted, but it's not clear to me why we do it this way anyway. Probably worth a quick discussion at some point. 

One more thing. Knox, there was some weird code in spectra.c which may have come about from an incorrect merge and didn't compile. I commented it out but it might be worth checking it still does what you want (sections using max_scat and so on). 

Generally, it would be good if someone, probably @kslong , could review these changes.


kslong:
Before we merge back to dev we will need to merge the current version of dev into the fu ori branch.

I am still looking into this, but
* It does not look like these files have been run through indent.py properly
* I have looked at spectra.c, and we should simply replace the version that is here with the version in dev.  As you say, there appears to have been a merging issue with it
* I have verified that our regression tests are OK at least if you compare with an old py82k versions (In py83 I have switched the atomic data files to standard80, and there are differences.)

kslong:
@jhmatthews  The normal fuori branch fails the regression tests at present.  In addition to the problem with spectrum.c there is a problem with rdchoice for reading in the variable having to do with shock heating (that gives a warning on compilation, and fails on running).

Either we need to do a selective merge to fix this so that we can compare the old and new approach with kpkts, or we need simply to go ahead an merge your version of the fuori branch into the regular one, and go from there.  

We also need to deal with the fact that dev branch is going to have to be merged in.  Perhaps you and I should chat.

jhmatthews:
I've fixed the indenting. This branch can be merged, I will deal with the merge from dev, it will be easy enough. 

The regression testing can then be done compared to dev because the normal runs should be unchanged. Then we just need to understand the other questions I have above, but in principle if the normal runs are unchanged it could be merged into dev even if there were outstanding questions about the shock heating code.

jhmatthews:
When I say this branch can be merged, i mean this pull request can be merged, i.e. merging jhmatthews/fuori into python/fuori.

PR #480, opened 2019-03-01
Updating Makefile to add -lrt when required

saultyevil:
So far I've tested this on RHEL 6.10 using gcc-4.4.7, Ubuntu 18.04 using gcc-8 and macOS using clang/gcc-4.2.1. They all seem to be working fine. I'll merge in if this passes Travis.


kslong:
I wondered why compilation was now working on the linux systems at STScI.  That’s great Ed.
Knox

PR #481, opened 2019-03-04
Reverb & Documentation Update

smangham:
This contains:
* An update to `setup_reverb.c` to use `rdchoice`.
* An update to `autogenerate_parameter_docs.py` to cope with `rdchoice`.
* All existing documentation set to have correct `rdchoice` options and filenames (several had been renamed by case changes).
* A basic `reverb.pf` example file.

Just waiting on Travis to decide whether or not this is safe.


kslong:
I have merged this into dev.

PR #489, opened 2019-03-20
Changes to Exit

saultyevil:
@kslong, here are some proposed changes to the Exit wrapper function you were interested in. I've added a version of error_summary which uses Log_parallel instead of Log and also added a check to see how many MPI processes are running. If only one process is running, then exit is called instead of MPI_Abort. This should avoid any MPI warnings when exiting with only one processor.

I've also added extra code to run Setup_Py_Dir if the atomic data isn't found as we briefly discussed. For now, the checks are done after reading in the atomic data filename. In theory, the code can be cut and pasted to main and modified slightly for when you want to incorporate Setup_Py_Dir and the yaml files.

There are some merge conflicts, but I can fix those if you think the changes to Exit are good!


kslong:
@saultyevil   This seems reasonable as far as I can tell.    Instead of utility.c, I would name this file something that is more indicative of what it is for e.g. xlog_util.c  to indicate that these are utilities associated with logging.  I did not understand why some of the .h files were needed in utility.c.  As we would like logging to be something we could use on another program completely, these should be removed if possible.

One thing I noticed that would be good to correct.  As you pointed out originally, when we exit on an error we should not be exiting with 0.  Perhaps we could check that when Exit is called the exit code is not 0, and if is, change it to something, e.g -1, issuing a warning that says something like:

Programmer warning: Exit should not be called with a zero return, please fix.  

or something to that effect.

As you may have noted, I have been working on python scripts to run a series of models and it would be good to be able to catch bad runs.    

Of course, this suggests that we should do a sweep and fix all Exit(0) to something.

Thanks for attending to this.
Knox

saultyevil:
The extra headers were me trying to be clever and write out the diag filename at exit. I can change that to make the Exit function less Python-centric.
 
I totally agree with you on non-zero error codes. I also think it would be beneficial to have something like an enumerator of possible errors, i.e.

```c
enum ERRORS
{
   UNKNOWN PARAMETER = 1,
   NUMERICAL_ERROR,
   MEM_ALLOC_ERROR
};

Exit (UNKNOWN_PARAMETER);
```

This would organise the different types of errors which are possible and we could probably write a script to parse out the different possible errors and their respective codes for documentation purposes or for your batch script.

Of course, like you said, this means doing a sweep of calls to Exit, which would take a while since there's ~230 of them, so it might be more sane to update Exit as we come across them.

kslong:
Once you are satisfied, Ed, I think you should go ahead and merge it into dev, and we can continue to work to improve it over time.  As long as Exit has an nonzero exit at the end that will capture most of my concerns.   

I also thought about enumerating error types, but two things were not obvious to me:  (a) what a sensible set of types really was, and (b) how exactly I would use them, given that we would have printed out the Error message.    If we cannot figure out the answer to both of this questions, we might as well use the stdio.h answers EXIT_SUCCESS, EXIT_FAILURE  which map to 0 and -1

kslong:
One additional thing that occurred to me that we may want to do in this regard, is to have a way to dump the wind file when we exit with an truly unexpected error.  We would want the windfile from the thread that generated the failure, and we would not want it to supplant the normal windfile that may have been created.  This could be triggered by the number that is sent with the Exit command.

PR #491, opened 2019-03-22
Ksl speedup

kslong:
I have changed the way the -p option works.  Now invoking the -p option with no follow-on simple sets up a logarithmicly increasing numbers of photons going from nphot/10 to nphot.  If you provide a range, say 2, then it will go from nphot/100 to nphot.

I made the change for two reasons (a) I found with the previous approach that there were times when it never actually got to phot max, and (b) I think this is really a more useful way to proceed where in every cycle the number of photons increases some.


saultyevil:
Your changes looks reasonable Knox, I'm happy for you to merge!

Issue #451, opened 2018-11-28
Various atomic data files have errors


kslong:
In issue  #448 which I plan to shortly close, @jhmatthews  noted that we have various errors being thrown by lines in various atomic datafiles that we are ignoring.  This is dangerous.  Someone needs to review what the errors are and either fix the data files or the code.

This is issue is likely related to several other issues #396 and #195.

kslong:
This issue begs the question of what atomic datafiles we want to fix, a point wich @Higginbottom has made in the past.  My guess is what should happen is that we should define a new set to be used, e.g standard82, h10standard82, etc that have error removed and leave the old ones as they are possibly hiding them somewhere so people don't stumble across them.  Comments on this would be appreciated.

Our regression tests generally use standar78, but presumably standard80, the latest non-macro atom  data in that series..  it produces the following errors, on one of the regression test models that I modified to use standard80:

![screenshot_878](https://user-images.githubusercontent.com/2530901/49320295-7517dc00-f4c6-11e8-854e-77b0ff271a95.png)

Note I ran this with the usual -np 3, so there are often 3x as many errors as there really would be with a single thread.

I took the regression tests .pf files and converted all of the standard78's to standard80 and ran a regression test comparing standard78 and standard80.  The output spectra look pretty similar, except at the Lyman edge where for most of the dataset get more Lyman continuum created in emission below 912.  The emitted spectra look quite similar in all cases.

kslong:
There is a related issue #156

jhmatthews:
Nick, any idea what we need to do here? The errors are quite confusing and we should really get rid of them, but I'm not sure why one ends up with these:

Get_atomic_data: more than one electron yield record for inner_cross
Could not interpret line 126 in file data/atomic80/ver_upsilon_linked.dat
No inner electron yield data for inner cross section

Any ideas what to do @Higginbottom ?

kslong:
@Higginbottom @jhmatthews 

These are the current errors associated with Keara's C345 model

[foo.txt](https://github.com/agnwinds/python/files/3027093/foo.txt)

Nick, it would be really helpful if you could comment on them.   As far as I can tell get_atomic_data does not read lines beginning with SCUPS and SCT, which appear to be from (according to the information in get_atomic data) Burgess and Tully 1992.   Does this mean they should be expunged from the data files, or alternatively does it mean we should treat them as comments.  When were they removed from Python?

Someone, remind me also how the inner yield electron data is used?

[...]

PR #494, opened 2019-04-02
Fixed a bug in #451 

Higginbottom:
A failure to match a collision strength spawned two more errors due to the following lines in the data file


aaaaaadam01:
haha those indents

PR #552, opened 2019-07-11
Changes to allow v0 flexibility for SV winds as non-advanced mode 

jhmatthews:
This change allows the user to choose whether the initial velocity for SV winds as a non-advanced mode. It can be either fixed, in which case it asks for a value, or sound_speed, in which case it asks for a multiple of the sound speed.

All example files updates so travis should pass. 

This involves two new keywords so could mess up people's runs. So someone should decide when the best time to merge is @kslong @smangham @saultyevil


saultyevil:
I'm happy to merge whenever. I'm already using this as an advanced parameter for my TDE models, and I have a small script which I can use to modify parameter names.

jhmatthews:
@Higginbottom @mandyhewitt @kslong @saultyevil @smangham - I'm merging this, but note that it affects all SV parameter files.

Issue #542, opened 2019-07-08
Replace Numerical Recipies with GSL


smangham:
Numerical recipes has an annoying license file and we can't redistribute it. We need to switch NR out for GSL.

Higginbottom:
Closed with pull #558

PR #558, opened 2019-07-15
Upstream dev

Higginbottom:
This is a pretty major bunch of changes, completely excising numerical recipes code and replacing it with gsl equivalents.
 
This closes issue #542 

I've updated the version number to 83c because there will be different results from 83b for almost all models. See the issue page for comparison plots.


Higginbottom:
Removal of numerical recipes

PR #559, opened 2019-07-16
Continued clean up of variable names

kslong:
In this case, the changes are associated with describing a power law for a boundary layer


kslong:
Well, I am now in a position to switch over the variables from BH.whatever to Central_Object.whatever.   This leaves Boundary_layer.whatever as a separate set of variables.

Issue #523, opened 2019-06-05
rtheta grid has angles > 90 degrees that are "inwind"


jhmatthews:
I've been fiddling around with rtheta grids for some spherical models and I'm noticing that I have cells that are classified as "in the wind" (i.e. inwind = 0) with angles > 90. There are two out of three cells with this theta>90 that have inwind == 0. 

I understand there need to be "ghost cells" in python, but I don't actually know why there are 3 (note there are also three in cylindrical coordinates). If this is correct, then I don't think we should be setting inwind to 0 for these cells, because they appear in things like convergence statistics and no photons ever get to them.

Also, the x and xcen values of these cells are not actually consistent if you inspect rtheta_make_grid (), and it's inconsistent with what's done in hydro (I think). So we should probably figure this out at some point.

kslong:
Can you provide an example of this.  I don't see it when I run the attached .pf file
[test.pf.txt](https://github.com/agnwinds/python/files/3363116/test.pf.txt)

jhmatthews:
Use case: if you run the attached model, then there appear to be cells on the outer edge of the grid that have very small z values:
[template.pf.txt](https://github.com/agnwinds/python/files/3378016/template.pf.txt)

These cells get no photons, but do have a density and contribute to convergence statistics. Thus, a fully converged model might get reported as not converged. They also seem to be included in the wind (i.e. their inwind values are >=0). 

The portion of the code I am referring to above, regarding inconsistency of x,xcen and theta,thetacen, is here (line 191 of rtheta.c):

```C
      /* Only the radial distance can be logarithmic */

      theta = w[n].theta = dtheta * j;
      thetacen = w[n].thetacen = w[n].theta + 0.5 * dtheta;
      if (theta > 90.)
      {
        Error("Angle greater than 90!!\n");
        theta = 90.;
      }
      if (thetacen > 90.)
      {
        thetacen = 90.;
      }


      /* Now calculate the positions of these points in the xz plane */
      theta /= RADIAN;
      thetacen /= RADIAN;
      w[n].x[1] = w[n].xcen[1] = 0.0;

      w[n].x[0] = w[n].r * sin (theta);
      w[n].x[2] = w[n].r * cos (theta);

      w[n].xcen[0] = w[n].rcen * sin (thetacen);
      w[n].xcen[2] = w[n].rcen * cos (thetacen);
```

kslong:
Here is a snapshot of one set of cells at a particular radius 
[...]

PR #562, opened 2019-07-18
Bug523 rtheta

kslong:
This addresses issues associated with #523 and ghost cells near the x-axis in rtheta models.  Specifically
- Ghost cells should now be properly labelled
- The number of ghost cells along the plane of the disk has been reduced to 2
- The angles for the theta grid has been altered so that there is no "real" cell beginning at 90 degrees.  Instead, the last real cell ends at 90 degrees.

### DIFF
--- a/source/cylind_var.c
+++ b/source/cylind_var.c
@@ -361,7 +361,7 @@ cylvar_wind_complete (ndom, w)
 
 /**********************************************************/
 /**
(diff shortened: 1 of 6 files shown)


jhmatthews:
This looks right to me Knox. I suggest we merge this and close the issue, at least until anything related resurfaces.

PR #570, opened 2019-07-31
option to avoid modifying prototypes with run_indent.py

jhmatthews:
Currently, if you run ```py_progs/run_indent.py -all```, or equivalently ```make indent``` on the source files, the indenting will show differences in templates.h, log.h and atomic_proto.h. This is a pain and shows unnecessary diffs in my opinion.

This proposed change adds an -all_no_headers option to run_indent to ensure that make prototypes is the only thing that modifies the prototypes. I think this is preferable. It also changes the makefile so that it defaults to using this option.

### DIFF
--- a/py_progs/run_indent.py
+++ b/py_progs/run_indent.py
@@ -148,17 +148,28 @@ def doit(filename='lines.c'):
 
 
 
-def do_all():
+def do_all(ignore_list=None):
     '''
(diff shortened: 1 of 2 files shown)


jhmatthews:
does this look right @kslong?

kslong:
Yes this seems very reasonable to me

Issue #571, opened 2019-08-02
zmax set incorrectly in rtheta import model


Higginbottom:
I've been trying to set up a simple shell model to test the flux calculations in python. All was going reasonably well (apart from the tricky nature of making an imported model) but when the thickness of the shell exceeded a certain value I get this flux vs angle (should be flat)

![Flux_by_bin](https://user-images.githubusercontent.com/3329213/62366546-1c01bb80-b51f-11e9-93e1-921c023447c5.png)

After some investigation, I discover that although the shell is 1e11cm thick, and has an inner edge at 1e11cm, zmax is being set to 1e11. rho_max is 1e12 and rmax is 1e12 as should be. Compled with the fact that smax ends up set to 0.5e11 what happens is that photons are forced to scatter midway through the shell - they go straight on which is fine (density is very low) but for those in more play bins, midway through the shell pushes their z component above zmax and so they are lost. Hence the factor of 2 drop for polar cells - the photons only do 50% of the path through the cell before they are deemed to be outside wind.

I see where this is done (in rtheta_make_grid_import in import_rtheta.c) - just working out the correct logic and then will fix.

Higginbottom:
So, the answer seems too simple - I'm concerned I've misunderstood the logic.

When we read in the grid - the i number is for the radial cell number, and the j is for theta cell number. Once the grid is read in, ndim is set to the number of radial cells, and mdim is set to the number of theta cells.

So, when we come to rtheta_make_grid_import we first loop over all the cells we have read in. The inner coordinate of each cell is set to w[nn].x[0],w[nn].x[1],w[nn].x[2].

There are now a couple of lines used in domains - which I think uses the wind_x variable to store the radial cell boundaries, and the wind_z variable to store the theta cell boundaries.

Then the maximum z,rho and r extent of the wind is computed. This is used to work out if photons have left the wind during transport. 

So we loop over all the cells once more. 
I think this loop is intended to loop over all the theta cells at a given radius, then the next theta shell out etc. 

so r_inner is initially just set to length (w[nn].x)

then nn_outer is set to nn + xx_rtheta.mdim - which is the cell one cell radially outwards.

now, if we aren't in the last radial shell checked by ( if (nn_outer < zdom[ndom].ndim2). )

we set r_outer to the length of w[nn_outer].x.

Ok, we now test if the cell is in the wind, if so the first check is

[...]

PR #582, opened 2019-08-08
zmax in imported rtheta models

kslong:
Update how various boundaries (e.g. zmax) are set up for imported
rtheta models.  See #571.


Higginbottom:
Seems to work OK in my test.....

PR #591, opened 2019-08-19
Remove local GSL

smangham:
Removed the local GSL.


smangham:
It looks like the line `#define C 2.997[...]` in `atomic.h` is causing problems with GSL, as it replaces *all* instances of a single capital `C` with `2.997[...]`, and GSL 2.6 adds a few variables named `C`.

So arguably this is a problem with us dumping a *lot* of very generically-named constants in the global namespace. [GSL provides its own physical constants](https://www.gnu.org/software/gsl/doc/html/const.html) but they have very long and clunky names.

I'm trying to reorder the imports to see if this fixes anything...

jhmatthews:
I suppose really we should have some generic prefix for constants. E.g. I think pluto deals with this by using const_C const_PI const_G.

smangham:
Yeah. We could create a struct called `pc` (for physical constants) and declare properties of it e.g. `pc.C` to make things short but it looks like consts in structs are awkward. Defines are probably just the best option, so I'll go through and rename the offending variables to `const_*`...

smangham:
It looks like the content in `my_linalg.h` is now included in GSL. This is potentially an actual problem rather than a nuisance like having to rename constants. What version of GSL is currently on Iridis? @saultyevil?

saultyevil:
On Iridis 5, it looks like I have access to GSL 2.4. But, Iridis 4 has access to GSL 2.6... I think I need to send an email to someone on the Iridis 5 team.

Higginbottom:
Yeah - I got them to install GSL2.6 last week on iridis4.

smangham:
Brilliant then, we can drop `my_linalg.h`.

kslong:
If we are going to go the route of removing gsl from the distribution then we cannot do anything that relies on a particular gsl version.  We should just rename C to something else, but not anything as nasty, or generic as what @jhmatthews proposed. 

Wikipedia says: 
Historically, the symbol V was used as an alternative symbol for the speed of light, introduced by James Clerk Maxwell in 1865. In 1894, Paul Drude redefined c with its modern meaning. ... Sometimes c is used for the speed of waves in any material medium, and c0 for the speed of light in vacuum.

It would be fun to use V, but maybe better to use C0 or possibly VLIGHT.  I think lowercase c will lead to disaster.

Note:  I did at one point have to rename one of the routines in Python due to a similar conflict.

smangham:
I'd advocate against `V` and `C0`. For future proofing, we should really avoid using short (1-3) macro names due to the high risk of collision with a future update to GSL. 

That said...

It looks like replacing the `#define` statements with `const` ones might actually be better- StackOverflow suggests that during compiler optimisation `const`s will be converted to static numbers, just during compile rather than preprocessing. You can't use them for things like declaring array sizes (so it wouldn't work for `NWAVE`) but for physical constants it'd be fine. This would be the easiest fix, as the `const C` will be stomped on in the local scope of the GSL function but be accessible without changing it everywhere else. We'll get warnings, but we already get warnings 🤷‍♂. 

I can give that a test and see if it works and if it gives a performance hit.

If it doesn't, I'd vote for @jhmatthews' idea of prepending `CONST_` to our define names, or maybe `DEF_` to make it absolutely clear that these things aren't variables or even actual constants. We could always only do it for those defines with a name 1-4 characters long, and keep things like `BALMER` as they are- but otherwise `E` and `H` are just going to cause this problem in another year or two.

kslong:
I will veto the CONST_ approach.  That may be a great approach for a computer scientist, but it's exactly the wrong thing for an astrophysics program.  

I also don't want to end up with ignored warnings, so we are going to need to figure this out in a way that does not end up with that.

As I indicated in a separate message, one way may be to isolate the gsl calls with an interface/isolation layer.  That just means putting all the gsl calls in a separate .c file and making sure we do not need to include python.h or atomic.h there.

smangham:
I think writing an isolation layer will obfuscate the code a bit. Switching `C` -> `VLIGHT` is a good enough fix I think? For future proofing, we should probably also switch 
* `H`-> `PLANCK`
* `E` -> `CELEC` (to match `MELEC`)
* `G` -> `GRAV`.

kslong:
CELEC, unlike melec or m_e  (to contrast with m_p) does not make sense.  

I would like to have a discussion of our whole approach here with others.  I'm concerned generally about whether this is a good idea, given the time it has taken up, and that we are now taking defensive measures for possible future changes in gsl.  This suggests we are anticipating more problems in the future (and this may well be correct.) I think we should ask the question whether we would not be better off simply distributing the version of gsl we want to do use.  (One reason by the way I think this might be better is for debugging problems.  If we know which version of gsl we are using, we can expect that for the same pf file, we will produce exactly the same photons.  At the point we use "random" versions of gsl we wil not be able to compare across platforms and users.

jhmatthews:
Hi Knox. Why can't we just specify what version of GSL to use? 

If the user wants to reproduce results exactly, they have to use a certain version. Whether the distribution actually *ships* with the code i.e. is in the folder or not, is basically immaterial, since that's just the action of copying some files somewhere. I also think that's a pretty standard software way of specifying "pre-requisites", and means less work for us since we don't have to actually maintain GSL in the repo - plus it prevents users having conflicts (don't actually know if this can happen or not, but it feels like it could if they use GSL elsewhere).

kslong:
Because they will not do that.  They will use whatever version  they have installed and then our first question will be what version they are using, and either we will have to say we don’t support whatever version they are using or we will have to work with them.  Basically, if we don’t fix the version and put into our installation (either virtually or actually), then it will be harder to support.  Note that it may be possible for us, rather than having gsl part of our distributin, to retrieve the version we support from gnu.org … though I don’t know that I see that being particularly better than including it as part of the distribution.

jhmatthews:
Hmm, well I don't really agree with that. I think it's standard (and good) practice to expect someone to keep a library up to date. We don't ship a compiler, or MPI, with the distro and similar things can happen there (e.g. MPI also deprecates functions and changes calls). We can even assert this on compilation if we really want, in the style of a python package install (using gsl-config in the configuration stage). 

Having said all that, I don't think this is a huge issue either way, so if someone really does want to keep it as is I don't really mind, it works, after all, so we should just move on and leave it as it is.

smangham:
So the problem with sorting Travis seems to be that Travis OSX default version of Python is 2.7, and our tests no longer work on Python 2.7. Trying to install both Python3 and GSL via Homehrew causes a mystery crash. I'm experimenting with different ways of installing Python3 and GSL.

kslong:
I have checked clean installs of the current dev Python (where there are no modifications to the constants, like C and H) on my mac and on our linux cluster.  I still want to have a reproducible error before we make any such changes.

@jhmatthews   You say it "standard and good practice" to expect someone to keep libraries up to date.  I have to disagree strongly with this statement, and I can give you several widely used places where that is not the case, beginning with conda and continuing with the data analysis systems associated with HST, Chandra, XMM, HEASOFT (which includes XSTAR).  All of these systems pick the versions of the external software they used, including typically the version of python.  Obviously, there are many times folks choose the opposite, but to me the choice depends on whether you think there are likely to be problems.  

(By the way, it seems to me, we have kind of lost wher we want to document this whole problem.)

PR #597, opened 2019-08-27
fixes for macro-atom estimator errors, closes #583

### LINKED ISSUE #583: mc estimators - stimulated emission
The following .pf file does not run even after I but a limit on the number of times a sobolev density error was found.

[flat_rwtri.pf.txt](https://github.com/agnwinds/python/files/3483954/flat_rwtri.pf.txt)

```
Para:        20 -- model: %s %8.2f %8.2f
Para:        20 -- model: Rescaling spectra because parameter %f outside bound %f of spectra in grid
Para:        14 -- photon_checks:   nphot  origin  nres  freq     freqmin    freqmax
Para:        42 -- photon_checks: %id %5d %5d %10.4e %10.4e %10.4e freq out of range
Para:       398 -- The net flow out of simple ion pool (%8.4e) > than the net flow in (%8.4e) in cell %d
Para:        99 -- sobolev: VERY BAD den_ion has negative density %g %g %g %g %g %g
Para:         2 -- walls: distance %g<0. Position %g %g %g 
Para:    100001 -- mc_estimator_normalise: bb stimulated correction factor is out of bound. Abort.
Para:    100000 -- stimfac %g, i %d, line[config[i].bbu_jump[j]].nconfigu %d
Para:         1 -- sobolev: suppressing negative density errors
```

This needs your attention @jhmatthews  Presumably it is the same problem that @saultyevil  has been finding with his TDE models, but this is a CV with a very flat wind profile.

jhmatthews:
This change should hopefully solve #583. A few main changes:
* Fix to a mistake in how the error check on the density floor for fractional matom populations was carried out in macro_gov.c
* only report population inversion if the upper level is above the density floor 
* same approach in sobolev() in resonate.c

without these mistakes, I don't get the errors in the parameter file Knox provided as part of #583. 

It's possible there's a slightly better way to do this, e.g. one could clean populations after one sets the floors, but this seems a reasonable enough approach to me for something that won't affect overall calculations at all.

@kslong @saultyevil you can either test before merge by checking out jhmatthews/estimator_fix, or can be checked after merging into dev. We probably want to test for same results, and for no more of the sobolev and estimators errors that were happening.

### DIFF
--- a/source/estimators.c
+++ b/source/estimators.c
@@ -429,7 +429,7 @@ mc_estimator_normalise (n)
 
 {
   double volume;
-  int i, j;
+  int i, j, nlev_upper;
   double stimfac, line_freq, stat_weight_ratio;
   double heat_contribution, lower_density, upper_density;
   WindPtr one;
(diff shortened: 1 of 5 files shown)


saultyevil:
Thanks for this James, I've queued my problem model onto Iridis with your branch. So should hopefully have an answer RE number of errors by the afternoon tomorrow!

saultyevil:
Well, the model no longer has `mc_estimators_normalize` or `stimfac` errors and the `sobolev: VERY BAD ...` errors are reduced. But the model still crashes because of `macro_pops: level %i has calculated pop. %8.4e in cell %i`. This is with CNO processed abundances, by the way.

jhmatthews:
Thanks Ed. Can you post an example of the sobolev errors? 

So it crashes because of too many of those macro_pops errors? That I suppose it is a separate issue and probably the code is struggling to find a solution for the macro-atom level populations in lots of regions.

saultyevil:
Sure, interestingly they only occur in the first 4 processes out all 200...

```
sobolev: VERY BAD den_ion has population inversion 7.76661e-09 7.76661e-09 5 3 3.33882e+12 0.01368
sobolev: VERY BAD den_ion has population inversion 7.76661e-09 7.76661e-09 5 3 1.6152e+14 0.00906
sobolev: VERY BAD den_ion has population inversion 7.76661e-09 7.76661e-09 15 9 6.32741e+13 0.01727
sobolev: VERY BAD den_ion has population inversion 7.76661e-09 7.76661e-09 5 3 1.38921e+14 0.00194
sobolev: VERY BAD den_ion has population inversion 7.76661e-09 7.76661e-09 9 3 7.9732e+13 0.01722
sobolev: VERY BAD den_ion has population inversion 1.50911e-08 7.76661e-09 3 1 7.60895e+14 0.00056
```

It looks like in a lot of cases that d1 and d2 are the same in most of these errors, but there are some where d1 != d2.

jhmatthews:
Ed has confirmed that the Errors disappear with this more sensible error checking. This means there are no real population inversions, only things that happen when the density floor is being hit. Thus I am merging this and closing the relevant issue.

Issue #584, opened 2019-08-09
Reposition() moving photons below accretion disc


saultyevil:
Currently in most of the spherical TDE models, a fraction of photons are undergoing resonant scatters very close to the surface of the accretion disc, say around z ~ 1e5-1e6 cm. Due to scattering, photons now point in the -ve direction. At of end the while photon transport loop in trans_phot_single(), a call to reposition() is made which moves the photon a distance dfudge to avoid situations where the photon interacts with the same resonance. But reposition has now placed the photon below the accretion disc. After this, a call to walls() is used to check if the photon hit the boundary. 

However, when the code attempts to calculate the distance s (line 849 in photon2d.c),

```c
// pold is above the disk
s = (-(pold->x[2])) / (pold->lmn[2]);
```

s is negative, as both x[2] and lmn[2] are positive, triggering the error "walls: distance %g<0. Position %g %g %g". walls then returns istat = -1 and so the transport of this photon is ended.

Here is a pf file which replicates this problem after around ~35,000 photons have been transported. 

[tde_spherical_walls.pf.txt](https://github.com/agnwinds/python/files/3485854/tde_spherical_walls.pf.txt)

saultyevil:
Sorry, but this is going to be a bit of a brain dump...

To summarise this problem.. disk photons are emitted and then interact near the very edge of the cell, i.e. say at z ~ 1e6 cm. It's possible for some photons to now point in the opposite direction, i.e. pointing in +ve z to -ve z. If the interaction was with a resonance line, then the photon is moved some push through distance dfudge, typically ~1e7 cm from my TDE models, which in some cases causes the photon to pass through the disc plane without the code being able to properly handle this situation.

For now, since this only seems to effect a small fraction of photons, the fix I've implemented in my fork is to catch this odd behaviour and tag these photons as `P_ERROR` and being lost to DFUDGE. Essentially saying that these photons have undergone some undefined behaviour.

[...]

PR #599, opened 2019-08-29
Fix for reposition bug

saultyevil:
This is the proposed fix for issue #584. Essentially, it treats the photons which are pushed into the disk by dfudge as a special case and deals with them by moving them a more _appropriate_ distance from the last resonance interaction. In principle, it could potentially be extended to help treat photons lost to dfudge as well. Results from the regression test suite look positive as do tests with my TDE models.

I also made a slight change to Exit to flush the logfile as sometimes it was not fully up to date on exiting.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,10 +19,12 @@ set(CMAKE_C_COMPILER mpicc)
 add_definitions(-D MPI_ON)
 
 # Include the source
-include_directories(source)
+include_directories(inc)
+link_directories(lib)
 
 # Define the executables
 add_executable(python
+        source/rdpar_init.c
         source/bb.c
         source/get_atomicdata.c
         source/python.c
@@ -70,7 +72,6 @@ add_executable(python
         source/matom.c
         source/estimators.c
         source/wind_sum.c
-        source/yso.c
         source/cylindrical.c
         source/rtheta.c
         source/spherical.c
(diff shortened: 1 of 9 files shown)


kslong:
Does this work if the disk is NOT flat?

saultyevil:
Ah, that completely slipped my mind. I'll test some vertically extended discs before I move any further.

saultyevil:
As far as I can tell, the fix I have implemented does not effect geometrically thick discs. It's pretty much coded so it only affects the DISK_FLAT checks in `walls ()`.

In the discs I did try there are photons which seem to become "stuck" in the disc between two translates as the error "walls: %d The previous position %11.4e %11.4e %11.4e was inside the disk, correcting by  %11.4e" is printed. These photons have not undergone a resonant scatter and so aren't being pushed into the disc by `reposition ()` so don't seem related to the original issue this fix is dealing with.

kslong:
That was sort of what I was worried about because I did see that things were specifically set up for a flat disk.  While I expect your fixes have “done no harm”, I just wanted to make sure that a similar problem did appear with a thick disk.  Running the same tde model that generated the original problem, but using a thick disk should tell us.

saultyevil:
I have double checked the same TDE model again and with a standard cv model, both with various parameter choices for the vertically extended disk. I have not seen any evidence of a photon passing through the disk plane like in the flat disk case due to reposition (), but at least in the TDE model something potentially curious is going on either way - long story short is that ds_to_disk is sometimes returning large values for shallow angles... But, this is something we can discuss in the telecon and decide if this is an issue or not.

Issue #619, opened 2019-10-17
Not everyone is indenting the code; we need a better solution


kslong:
We are still having trouble indenting the code consistently because not everyone runs 
```
make all
```
as a standard way to compile.  Instead, we run
```
make python
```
if we are changing something in one of the python subroutines which does not automatically indent changed code.  As a result the right person is not getting credit for code that changes that he or she has done.

The solution is to run indent.py after all of the basic statements ... assuming this does not cause some other problem (e.g with make)

kslong:
A change to the Makefile which causes indent to be run when python, py_wind, or windsave2table are compiled has been implemented and merged into dev.  This should close this issue

PR #620, opened 2019-10-17
First attempt to create a Makefile to indent each time one of the rou…

kslong:
…tines is compiled.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,8 +59,8 @@ script:
     - echo $PWD
     - cd source/
     - make clean
-    - make CC=gcc py_wind
-    - make CC=gcc windsave2table
+    - make CC=gcc INDENT=no py_wind
+    - make CC=gcc INDENT=no windsave2table
     - make clean
     - cd ../examples/travis/
     - Setup_Py_Dir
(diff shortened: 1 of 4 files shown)


kslong:
@jhmatthews @smangham   I have made modifications to the Makefile and to the travis.yaml file that are intended to address #619  which was needed because I was finding that folks are not always compiling with make all, before pushing, and hence I was still seeing a lots of cases where someone's changes were not properly indented.

I have something that works now, although I don't really like it, particularly because make all now runs indent.py 3 times.  (It's fixed so it only does this on files that have been changed and not committed, but still ...)  

I would have preferred to have put something like

all: clean python py_wind windsave2table indent

and similarly for making python

xpython:  what we now have for pytthon
python: xpython indent

But I could not get this to work.  

Furthemore, to get travis to run, I had to add an option to the Makefile so that indent was not run there.  If I try to run indent.py on travis, it says that numpy is not found ... which is to me quite bizarre as I can see where numpy is loaded in the travis configuration.  I would prefer it, if indent was run ... though as this is likely not a git directory it might not work any way.

Anyway, could you  take a look at this and see if it makes sense, or rather if there is anything that is obviously somewhat better to do.

PR #656, opened 2020-01-03
Bug643 atomic

kslong:
James, Nick  - Could you look at this branch which puts the atomic data directories in with Python, and the models grids of spectra into a new repository xmod.  I would like to merge it into dev early next week.  

The main change from an operational perspective should be that the master files for atomic data now have .dat attached to them.   

Please look at the installations instructions (which tell you how to get the  spectral grids).  It should no longer be necessary to download the data repository, but it should also be true that it is not harmful to have the old data directory there.

Note that there would be an issue running this version of python with files that are unique to the old data directory, but only because Setup_Py_Dir now links to the directory xdata (rather than data).  If you go back to old versions of Python, before this change then Setup_Py_Dir would look at the old directory.


jhmatthews:
This looks reasonable to me Knox and worked for me. I couldn't check that all the new files were correct since there were some genuinely new files, but I can't really foresee any issues.

Issue #652, opened 2019-12-06
macro_gov (and related routines)


kslong:
Macro-gov is poorly commented and this needs to be improved, I have done a little (latest version of dev), but it would be better if someone who is more involved does this correctly..  
- Furthermore, it seems to return either 1 or 0, but this is not documented, nor is the reason for it identified. The returns are not used as far as I can tell.  
- Furthermore, it contains code that is impossible to get to, see below.  
- Additionally, it's not obvious, why one should be adding reverberation path lengths as is done below for the purposes of reverb.  macro_gov is not about moving photons from place to place.   - Also, the code is set up as a big while loop, but at various places we change the variable escape to 1 (which escapes the loop), and at other places we call return directly.    We should be consistent.   
- Finally the code operates very differently during spectral generation, but instead of dealing with this generically, a lot of code gets repeated (and I could imagine there is a more efficient way to do this, that would make the code more obvious during ionization cycles.)

Given our current issues with #648, we should clean this up.  

(There also smaller issues with the desciption of the routine macro_pops in the same file.  

```
          matom (p, nres, &escape);

          if (escape == 1)
          {
            /* It is going to escape as a r-packet that was created by de-activation of a macro atom.
               Therefore, if the frequency is suitable it should be recorded as a macro atom emission event for use in the
               computation of the k-packet emissivity needed for the final spectrum calculation. */
            *which_out = 1;

            //SWM - If reverb is on, and this is the last ionisation cycle, then track the photon path
            if (geo.reverb == REV_MATOM && geo.ioniz_or_extract && geo.fraction_converged > geo.reverb_fraction_converged)
            {
              line_paths_add_phot (&(wmain[p->grid]), p, nres);
            }
            return (1);
            /* 0803 - ksl - 60 - Added code to modify the photon origin to indicate the packet has been processed
               by a macro atom */
            if (p->origin < 10)
              p->origin += 10;
            return (1);
          }
        }

```

kslong:
matom.c is another matom routine where the Doxygen commen ts do not follow the satandard approach.

kslong:
estimators.c has some functions that have non-standard dox headers. This also needs fixing.

jhmatthews:
Had a look a this, see pull #657 

Some comments:
* Commenting: looked pretty good to me, but I fixed a few places where doxygen format was missing and cleaned up a little bit of formatting.  
* Impossible code: I removed the first return in the above code snippet
[...]

Issue #643, opened 2019-11-20
Mistake in CIV luminosity in py_wind? [diagnostic only]


jhmatthews:
I think I've discovered an error in how CIV line luminosity is reported in py_wind. Pressing "L" allows you to get the line luminosity for certain lines, but this seems to be too high when I try for CIV for my AGN model - in fact the cumulative total is higher than the system luminoisity.

The bit of code is:
```C
if (lin_ptr[nline]->macro_info == 1)
        {                         //If this is a matom line
          d2 = den_config (&plasmamain[nplasma], lin_ptr[nline]->nconfigu);
        }
        else
        {                         //If this is not a matom line
          two_level_atom (lin_ptr[nline], &plasmamain[nplasma], &d1, &d2);
        }
        x = (d2) * a21 (lin_ptr[nline]) * H * lin_ptr[nline]->freq * w[n].vol;

        if (geo.line_mode != 4)
        {
          x *= z = scattering_fraction (lin_ptr[nline], &plasmamain[nplasma]);
        }
```
which looks different to the equivalent function in lines.c. I would have thought a simpler approach is simply to replace the above lines of code with something like:

```C
    x = lum_lines(&w[n], nline, nline + 1);
```
which will be consistent with the normal operation of the code.

This might mean that there is an error in Fig. 8 of Mangham et al. 2017 (first reverb paper) where we show CIV emissivity, and possibly also in Fig. 2 of my 2016 paper, although it looks like the code was changed in between those two papers.

jhmatthews:
I can fix this issue easily, but do we need to do anything about past publications? (see comment above)

kslong:
We should definitely fix this error so it does not propagate.  And we should track down when py_wind changed to make sure we know what papers were actually affected.

jhmatthews:
This change seems to have been made in commit c085393ea on 2017-06-16 - but this involves a commit called "Updating dev" so it may have been made earlier in reality. This places it during the review stage for the first reverberation paper so I'm guessing that the emissivity plots for CIV are not quite right, but that the figures in my 2016 paper are safe. Not sure what we want to do about that @smangham

jhmatthews:
[...]

PR #657, opened 2020-01-16
Cleanup for macro routines

jhmatthews:
Mostly commenting.

Includes a few minor code changes, e.g., removing the double return statement in macro_gov() and ensuring that return is 0 not 1. 

See #652


jhmatthews:
Now also includes a fix in py_wind_ion.c with issue #643 to get the line luminosity result correct.

Issue #667, opened 2020-02-18
imported models - only T_e is set


kslong:
For imported models, we have the option of setting the a temperature.  But this temperature is only appled to T_e, and not T_r.  Both should be initialize, possibly using Leon's old approximation that T_e was 0.9 T_r.

saultyevil:
In one of my test runs of Jane Dai's model, I have in the model grid, dai.grid.txt, a run of gas temperatures provided to us. The function which reads in an r-theta grid has no trouble reading in the grid, but the temperature I have provided does not seem to be set.

In the provided parameter file (which doesn't all make sense, as I was just testing what the grid looks like), I do not run any ionization cycles but go straight to the spectral cycles so I can get some optical depth diagnostics - but I quit the program before I run any spectral cycles anyway.

When I run windsave2table after doing this and plot the snapshot, I get something looking like this, where the electron and radiation temperatures is set to a single temperature throughout the grid.

![image](https://user-images.githubusercontent.com/19627279/75260568-b8541e00-57e1-11ea-9b8b-715f526b9ab5.png)

As far as I can tell, whilst we read in the temperature for each cell and store it in `xx_rtheta.t[icell]`, this variable is never used again and only has scope in the file `import_rtheta.c`. I think this is the case for cylindrical models as well, but for the struct `xx_cyl.t`.

Incidentally, the temperature in the above plots is ~10^4.6 ~ 40,000 K which is the value of Wind.t.init. It looks like the colour bar has a slightly different scale for the radiation temperature, so I assume we are actually initialising T_e ~ 0.9 T_r in this case.

My guess is that this is to do with this if in define_wind() in wind2d.c line 339.

```c
if (zdom[ndom].wind_type == HYDRO)
{
    plasmamain[n].t_r = hydro_temp (x);       
}
else
{
    plasmamain[n].t_r = zdom[ndom].twind;
}
```

It looks like t_r is initialised to zdom[ndom].twind (or Wind.t.init) unless we're using a hydro model, where it is then set to whatever hydro_temp returns for the given position x.

jhmatthews:
Just a thought Ed, but we should probably avoid sharing the model grid on github yet - maybe put it in a privately accessible place for the moment.

saultyevil:
[...]

PR #670, opened 2020-03-03
Importing electron and radiation temperatures

saultyevil:
This PR is to fix issue #667. Both T_r and T_e can now be set with an imported grid. The behaviour of the how models are read in can be found in docs/sphinx/source/import.rst, I think I've detailed the majority of the information required to create a model to import. 

To quickly give an example of how reading the temperatures work now, for a spherical model we require the following columns,

i, r, v_r, rho

Optionally you can also provide the electron and radiation temperature, with the following behaviour,

* If no temperature is provided, T_e and T_r as set as Wind.t.init
* If T_e is provided, this value is set in plasmamain and T_r is set as T_r = 1.1 T_e
* If T_e and T_r are both provided, then plasmamain uses the values of T_e and T_r which were read in

In other news I've also "revamped" the routines in general. I've improved the documentation where I thought things weren't as clear, added in some safety nets (i.e. adding a catch to stop reading in when ncell > NDIM_MAX), and switched to dynamic memory allocation for the imported models structures (to be more memory efficient). I've unfortunately touched more files than I've actually modified, this has been due to fixing spelling mistakes whilst debugging and indent touching lines in other files.


saultyevil:
Does Travis still work for PRs?

kslong:
This looks great to me.  From my perspective, we should go ahead an squash and merge into dev.

I am not worried about the multiple domain issue, if it still exists, but it would be sensible to have Python throw and error if some tries to do this, and then exit

saultyevil:
I've fixed the multiple domains issue and tested by importing two cv_standard winds with different temperatures. In py_wind, each domain had a different temperature.

jhmatthews:
Weird, @saultyevil - indeed it seems travis isn't building PRs anymore. Seems it stopped a few PRs back, but I've no idea why.

kslong:
What are PRs and what does this mean for us?

Sent from my iPad

On Mar 3, 2020, at 10:59 AM, James Matthews <notifications@github.com> wrote:

﻿

Weird, @saultyevil<https://github.com/saultyevil> - indeed it seems travis isn't building PRs anymore. Seems it stopped a few PRs back, but I've no idea why.

saultyevil:
PRs = Pull requests.

It looks like Travis hasn't run for the latest commit in dev either..

Issue #686, opened 2020-03-30
1d wind model failing with errors


Higginbottom:
I am attempting to import 1D Pluto models into python in order to perform rad-hydro calculations.
My initial attempt failed with many errors of the type: spherical: ds)in_cell s and max returning VERY_BIG in cell ... and eve more translate_in_wind: Grid cell not in wind - we get a million of these, and the code crashes.

Following KSLs suggestions - I have simplified the situation for testing:

1: I generate a 1D model using the following parameter file:

[mod.pf.txt](https://github.com/agnwinds/python/files/4403689/mod.pf.txt)

2: This runs fine, and produces a wind save file. This is then processed through windsave2table.

3: We then process this through import_1d.py mod which makes a model file:

[mod.import.txt](https://github.com/agnwinds/python/files/4403704/mod.import.txt)

This has one inner guard cell (with zero density) and two outer guard cells - I assume this is correct..

4: We then read *that* model into python and run it via this parameter file. It crashes.

[import.pf.txt](https://github.com/agnwinds/python/files/4403722/import.pf.txt)

concerningly, the runtime errors look like:

ds_in_cell s and smax returning VERY_BIG in cell 1397164432 nudging photon by DFUDGE

which seems to have a dodgy plasma cell number....

Anyway - confirmation from others that their installation does the same would be appreciated. Plus any input from Ed..

Higginbottom:
Possibly related to #608

saultyevil:
Yep, I have the same problem as you. I've got the same problem in the latest commit of dev ([508233c
](https://github.com/agnwinds/python/commit/508233cc934c40f3cee275a44e3e50b851e6374a)) and a version of dev before I messed around with the import code  ([fd84d4d](https://github.com/agnwinds/python/commit/fd84d4dfb1b0a834d0b4ac5c3537270bf5d4d7ae))

I think the problem is something to do with this Error,

```
Error: translate_in_wind: Grid cell 28 of photon is not in wind, moving photon 8.51e+11
```

I do actually think it's correct that cell 28 isn't in the wind. But I don't know why photons are going into that cell. If I reduce wind.radmax to 5e12, then the model runs without throwing up any errors. But that's clearly the wrong value of wind.radmax...

I think the problem is in spherical_make_grid_import(). zdom[ndom].rmax = r[ncell - 1] when it should be r[ncell - 2], as there are TWO guard cells. Changing that value seemed to get rid of all of the errors.

[...]

PR #687, opened 2020-03-30
Change rmax setting to 1st guard cell rather than 2nd

saultyevil:
This should be the change to stop 1D models from crashing. The problem was that zdom[ndom].rmax was being set incorrectly, so photons were able to go into the final cell which had inwind < 0, but radmax had a value larger than the radial coordinate of this cell. This created a bit of confusion and a lot of errors being thrown.

I haven't changed the behavior of Python reading in Wind.radmax as mentioned in Issue #686 , as it turned out to be more complex than a quick fix. It looks like some internal checks require the maximum wind radius to be set to something, but for imported models we do not set the maximum wind radius before these internal checks.

Edit
The regression checks seemed to be fine...

### DIFF
--- a/source/import_spherical.c
+++ b/source/import_spherical.c
@@ -178,7 +178,7 @@ spherical_make_grid_import (w, ndom)
 
   zdom[ndom].wind_rho_min = zdom[ndom].rho_min = 0;
   zdom[ndom].rmin = imported_model[ndom].r[0];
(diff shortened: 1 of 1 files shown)


lazygun37:
Hey Ed,

Just thought I’d check in to make sure you’re doing ok? Must be at a least a little scary - hope things are getting better, not worse.

Stay safe and obviously don’t hesitate to call 111. And anything I can do - or you can’t - just let me know.

Cheers,

C

PR #690, opened 2020-04-08
Updating reverb documentation

smangham:
Updated reverb example, slight tweak to reverb.pf to reduce spectrum range. 

**Note:** In the previous examples, and the simulations I ran using them, I changed the `Wind.mdot` parameter and not the `Disk.mdot` parameter, which means that the data for the response maps is a bit wrong. I'm going to try and re-run these on whatever machine I still have access to.


jhmatthews:
Thanks Sam - We can merge @kslong @smangham ?

kslong:
Yes, I think we should go ahead and merge.  There is no downside to doing that.

PR #696, opened 2020-04-24
Fixes for imported rtheta boundary issues

saultyevil:
Bug fix for imported r-theta models.

I made a few typos because of confusing variable names and I was accidentally also using degrees instead of radians for sin() and cos().


Higginbottom:
Merged it - hope thats OK!

Issue #694, opened 2020-04-23
DFUDGE in spherical shell models


Higginbottom:
This is an old problem, that I'm bringing up again coz I spent a day boggling over it.

I have been trying to understand some aspects of flux logging in a 1D model in python, and as part of this I dis-interred some old tests. And got odd results, until I remembered that I'd had to hack python to get it to work.

Here is a spherical model - it is a thin shell..

[model.txt](https://github.com/agnwinds/python/files/4523753/model.txt)

if we run this in python, a59922445f878a1ffe0bbc262432b09c54e2dde9 

[shell2.pf.txt](https://github.com/agnwinds/python/files/4523759/shell2.pf.txt)

then we can obtain a value for the flux in the cells. We can calculate this by hand, and here is a plot of the flux vs theta for py84c (labelled standard) with the by hand flux as the black line.

It is low, and the reason for this is that DFUDGE is automatically set to 1e5 for this model. The shell is 1e6 thick, so photons miss 10% of the cell. If I hack python to set DFUDGE to 1, then I get the other line

![Flux_by_bin](https://user-images.githubusercontent.com/3329213/80122994-30428880-8586-11ea-9b7f-23e3cef12733.png)

Is this a problem in normal models? Not sure, although the reason I was digging was to look at a problem in my stellar wind models - so it might be.

However, it does point to an issue in automatically calculating DFUDGE, which has caused a lot of problems in the past. 

I think the issue causing DFUDGE to be large, is that the minimum radius used to calculate it might be being set to 0.0 because there are cells with a cartesian coordinate x=0.0. Should we use a radial measure in a spherical model. Not sure, just want to make sure we discuss this and agree on wether to work on it.

kslong:
So just to be clear,  Nick, this is a 2d model not a 1d model of a spherical shell, which is what I assumed originally. So whatever one chooses DFUDGE to be it must apply to all spherical polar models, whether they are actually spherical or not.  This shell has a thickness of 1e6 cm,  placed a a distance of 1e11 cm.  

I do agree however that something is odd:

For imported models rmin is set in import_rtheta.  It looks like it is setting rmin to the minimum radius ... but it looks like rmin is being set for any cell, and not restricting things to cells that are in the wind.  This looks like a mistake.

In your example, you have cells that are not in the wind with a size of 10e10, but your first real cell is at 1e11.  This means the program is finding that the thickness of the wind is about 1e11.

[...]

PR #699, opened 2020-04-29
Upstream dev

Higginbottom:
This is a pull to try and add a test to make sure DFUDGE is always smaller (1% at least) than the smallest grid dimension.


Higginbottom:
A pull to do more work regarding bug #694 - it also contains a fix to a minor bug in the calculation of flux - this had already been fixed in the version used in the HK20 paper. It also includes a new routine rad_hydro_files - that generates the files required to connect python to Pluto.

Issue #665, opened 2020-02-16
nioniz


kslong:
nioniz is said to be the number of photon passages by photons capable ion ionizing H.

In actuality it is the number times a wind_update_banded_estimators is called with a ionizing photon, but this can be many times the number of times an ionizing photon actually passes through a cell, since this depends on many factors.

nioniz is just diagnostic and not used for anything, but I think it would be better to count a photon only once. This is a better statistic for characterizing the number of photons that are ionizing in a cell.  

@Higginbottom  Do you agree?  

I think if you store the cell number and the photon number, and increment only if one of these has changed when wind_update_banded_estimators  is called, then we would get something that followed my proposed redefinition.

kslong:
There are other totals, e.g ntot, ntot_star, etc that also get summed every time you call radiation that have this property that probably should be fixed as well.  One would need to be careful that none of these are not used to renormalize some other quantity, assuming we redefine these. 

A suprise is that nioniz is updated in a different routine.  That seems a mistake.

Higginbottom:
I believe the main 'clean-up' has actually been done at some point - probably commit #666. Investigating the ntot logging...

Higginbottom:
I agree than nioiz and the other photon counters may as well all be updated in the same place. The same test that was put in place in #666 can also be used to only update the other counters once (although two new external variables are needed). I don't think they are used anywhere else to renormalise.

Higginbottom:
fixed with pull #703

PR #703, opened 2020-05-12
Upstream dev

Higginbottom:


Higginbottom:
This is a pull request to address problems with using the spherical model to compute flux - special modes created to log the radial flux in the first element of the various flux arrays, and also the forces. Also properly implemented the free free force. Also addressed bug #665

Higginbottom:
Oh, and also increased the version number since this pull affects the size of some arrays in the wind save and so 84e windsaves different from 84d...

Higginbottom:
Regression checks are identical, apart from rtheta and star - which seem to be due changes to DFUDGE which have been pulled into my branch. These are tiny changes though..

Issue #524, opened 2019-06-07
Surface reflection treatment: extract & live.or.die differ at high inclination


jhmatthews:
**Note: title updated. Originally a question based on misunderstanding but possibly still an issue to do with differences between extract and live or die.** 

I'm not yet sure if this is a problem but I think it's worth flagging. My understanding was that if I selected the "reflect" option for surface illumination treatment, then photons were perfectly reflected from the surface. If this is the case, then for an illustrative example with no wind, an isotropic SED with a reflecting disc (but no disc photons) should produce the same result as an isotropic SED without any disc surface at all, because a reflection is equivalent to travelling straight on into the opposite hemisphere (right?). 

Running this test reveals a flux difference in the case with the reflecting disc:
![nowind1](https://user-images.githubusercontent.com/3007249/59111273-9dcad380-8938-11e9-9553-9f4058cfc405.png)

whereas without a disc, all angles look identical:
![nowind](https://user-images.githubusercontent.com/3007249/59111271-9c011000-8938-11e9-821b-084dab2dd158.png)

To run this test, use this pf file and change disc type to "flat" or "none". 
[nowind.pf.txt](https://github.com/agnwinds/python/files/3266316/nowind.pf.txt)

kslong:
Sounds like a problem.  With a reflecting disk the photon actually goes straight through I believe.

jhmatthews:
The problem goes away when I switch to live or die, so this appears to be a problem about extracting photons that hit surfaces. 
EDIT: this was a mistake, see comment below.

kslong:
That sounds possible to me.  I don't remember ever thinking about that part of the problem, either for this case, or for the case where the photon is supposed to scatter at that point.

jhmatthews:
It looks to me like the issue might be in these lines of trans_phot.c (in this case line 373):

```C
      if (geo.absorb_reflect == BACK_RAD_SCATTER)
      {
        /* If we got here, the a new photon direction needs to be defined that will cause the photon
         * to continue in the wind.  Since this is effectively a scattering event we also have to
         * extract a photon to consturct the detailed spectrum
         */
        randvcos (pp.lmn, normal);
        stuff_phot (&pp, p);
        tau_scat = -log (1. - random_number (0.0, 1.0));
        istat = pp.istat = P_INWIND;
        tau = 0;
        if (iextract)
        {
          stuff_phot (&pp, &pextract);
          extract (w, &pextract, PTYPE_DISK);
        }
[...]

PR #715, opened 2020-05-30
Bug524 live

kslong:
@jhmatthews  Once you agree that this is correct, then this can be squashed and  merged.  I believe it address #524


jhmatthews (approved):

jhmatthews:
I think this can be merged after our discussion yesterday Knox. We could add an Error message if the user requests 90 degrees as an angle in live or die mode.

PR #731, opened 2020-06-16
Cleanup if statements regarding photon istat

jhmatthews:
This replaces a bunch of statements with a loop over photon istats. I find this cleaner code and easier to read, but not sure if everyone agrees. Probably other areas of the code that are similar.


jhmatthews:
This produces identical output files in terms of numbers (one change to text that's printed).

PR #733, opened 2020-06-17
Merge into master for version 84e

jhmatthews:


jhmatthews:
Probably there should be release notes associated with this written on the releases page.

PR #737, opened 2020-06-21
Bug736 disk

kslong:
This is intended to solve the problems @saultyevil  raised reagarding what happens when photons hit the disk.  

Basically in the new version, when photons hit the disk they are transformed to the local frame of the disk, redirected, and then transformed back.  (All of the transformations now involve the freq, not freq_orig, or freq_orig_loc, so this should be cleaner as well)

The changes don't modify the regression tests much.


saultyevil:
I think this looks good! It deals with the problem I was having and the question I raised. I'm going to merge this in so I can hopefully get some spectral cycles run before the telecon today!

saultyevil (approved):

Issue #749, opened 2020-08-03
Incorrect degeneracy in fully ionised simple atom _elem_ion files for metals


mandyhewitt:
Whilst working with the Fe K-alpha atomic data additions, Rahul spotted a potential issue in the macro atom files. In xdata/atomic_macro we believe that the fully ionised simple atoms are listed in the *_elem_ions.dat files with a degeneracy of 2 yet they should possibly be 1.

For example see the following in python/xdata/atomic_macro/CIIICIVCV_c10_CV1LVL_elem_ions.dat
around line 110

IonV   Li   3   1   2    5.39200 1000   0     1s^22s(2S_{1/2}) 
IonV   Li   3   2   1   75.64100 1000   0     1s^2(1S_0) 
IonV   Li   3   3   2  122.45500 1000   0     1s(2S_{1/2}) 
IonV   Li   3   4   2 1.0000e+20   0   0     Bare 
-----------^-----------------
degeneracy is shown here as a 2.

Line 110  (the last line above) should be:
 
IonV   Li   3   4   1 1.0000e+20   0   0     Bare 

The same issue can be seen throughout this file and in the other .elem.ions.dat files

This could be an historical issue which may have a significant impact on TDE models.

Higginbottom:
This was changed from 1 years ago - at the time of doing it, we all agreed it was right to be 2, and I think it solved some problems I was having - i.e. improved accuarcy.
not saying its definately 2 not 1 - but do you have a reference? I’ll try and dig up my notes - about 9 years ago….

Nick

Higginbottom:
LOL!!!
Found my notes
I changed the multiplicities from 2 to 1 back in October 2012 in the basic elem_ions_ver_80.dat - that is still correct.

So you are exactly right - the macro atom files should also have 1 for the bare ion multiplicity.

I didn't know anything about macro atoms back then - and indeed nobody was using the files so I didnt check..

Well done Rahul!

Deleted my earlier incorrect reply. Apologies for any confusion

Nick

jhmatthews:
We should check where this gets used - if it does, could feasibly affect my quasar runs as well. Looks to me like it's used in the recomb emissivity calculation at least. Bit confused as to how it happened because I thought those files were basically adapted from the standard ones. 

in future, might want to have a sanity check for bare ions, something along the lines of 
```C
if (ion[nions].g != 1 && ion[nions].istate == (ion[nions].z + 1))
[...]

PR #754, opened 2020-08-06
Updated degeneracy of fully ionised atoms in _elem_ions files and introduced check in atomicdata.c

rgupta106:
Corrected the degeneracy of the fully ionised states of atoms in _elem_ions files from 2 to 1 as per bug #749. Added a sanity check in atomicdata.c to check against this issue and if it doesn't pass then it changes the value of the degeneracy of the fully ionised state to 1.


jhmatthews:
Might be good for @kslong to check this, but it looks good to me.

jhmatthews:
We should merge this - you happy @kslong ?

PR #755, opened 2020-08-06
moved simple estimators to separate routine

jhmatthews:
I've moved the normalisation of the simple estimators to a separate subroutine and added a few XFRAME comments. @Higginbottom @kslong @saultyevil tagging you so we don't duplicate effort. Might be good for someone to check what I've done quickly.

### DIFF
--- a/source/estimators_simple.c
+++ b/source/estimators_simple.c
@@ -428,3 +428,128 @@ update_force_estimators (xplasma, p, phot_mid, ds, w_ave, ndom, z, frac_ff, frac
 
   return 0;
 }
+
+
+/**********************************************************/
+/**
+ * @brief Normalise simple estimators 
+ *
+ * @param[in] PlasmaPtr xplasma  PlasmaPtr for the cell of interest
+ *
+ * @return Always return 0
+ *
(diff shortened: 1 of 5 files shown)


saultyevil:
I think there could be a problem with the normalization of the flux estimators at the very bottom of normalize_simple_estimators.

I think we want the flux estimators in the observer frame, but the estimator is being divided by the volume from wmain which is in the CMF. Am I right, @Higginbottom ? I think we're missing a factor of gamma...

jhmatthews:
I think all the estimators here need a review on the volume here - I actually think there's a gamma missing in the other estimators here, because they need 1/Delta V Delta t on the bottom, and at the moment they have effectively Delta V_cmf * Delta t_obs. I've got an XFRAME comment in there so we don't forget, but yes deffo issues there.

Higginbottom:
That sounds right - where is this function. I just pulled from upstream belfast20, the most recent commit is 7429e82fa24b4554e59012496dd110b0e9bcba48
from wednesday…
tried to grep for that function - not there….

Nick

jhmatthews:
Yep, I haven't merged it yet so it's not in upstream. Can do so now if people are happy.

PR #762, opened 2020-08-28
added files for implementing fe as a macro atom to python

rgupta106:
Included are the files required for the implementation of iron (Fe) as a macro atom to Python developed during my 2020 summer project.


ssim:
This PR edits the "make_macro" script and adds (i) new atomic data for Fe XXV and Fe XXVI and (ii) adds an example .pf file for a simple X-ray wind model.

I think we can add it in ok.

PR #769, opened 2020-09-15
proposed change to distinguish between energies and luminosities for wind photon generation

jhmatthews:
This is a proposed change regarding emissivities. Specifically:
* make things like geo.f_tot to be CMF energy = CMF luminosity * delta_t_cmf
* keep the lum_ fields as co-moving luminosity

Otherwise I think we end up overwriting lum_ fields with different quantities in the luminosity and cooling routines, and also we are a bit inconsistent with our approach to other rates. 

actual changes:
* revert to old behaviour in the wind_luminosity routine (i.e. remove the Gammas so the lums are emissivity * CMF volume with no time correction)
* ensure that all fields like geo.f_wind have a gamma converting them to using Delta_t_obs (i.e., they are an energy used to generate photons)
* add a gamma in the wind photon generation routine where we use a lum_ field to work out which cell the photon comes from
* this requires one to be able to call wind_luminosity with two modes, some where the luminosities calculated as energies (luminosity * dt_cmf) for photon generation and others where luminosities are really CMF luminosities. 
* In **all** cases, the quantities in geo and xplasma always refer to CMF luminosities, sometimes summed over cells (and so not in one specific frame). 

I believe this approach will mean minimum confusion in terms of what the fields containing without needing a ton of gamma factors all over the place. It will also mean that fields containing luminosities really are CMF luminosities, although the summed quantities will be a little odd - but they would also be a little odd if they were CMF energy/observer time.

### DIFF
--- a/source/emission.c
+++ b/source/emission.c
@@ -22,8 +22,9 @@
  * @brief      calculate the luminosity of the entire
  * wind between freqencies f1 and f2
  *
- * @param [in out] double  f1   The minimum frequency 
- * @param [in out] double  f2   The maximum frequency
+ * @param [in out] double  f1         The minimum frequency 
+ * @param [in out] double  f2         The maximum frequency
+ * @param [in]     int     mode       integer telling us whether we want an 0: energy (lum * dt_cmf) or 1: luminosity
  * @return     The luminosity of the entire wind
  *
  * @details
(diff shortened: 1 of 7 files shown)


jhmatthews:
Merging this as per others suggestions, but use with caution!

PR #784, opened 2020-10-24
Bug782 div v

kslong:
@saultyevil  @jhmatthews    These changes are intended to address problems with calculating the velocity gradient tensor and div v calculations which Ed uncovered with TDE model.   The changes basically are intended to

- Improve the calculation of the derivative by adopting a symmetric +- ds scheme
- Properly calculating where to get the velocity at +-ds, by calculating the 3-vector in the observer frame that represents ds.  This is needed because although we want to calculate the velocity gradient and divergence in the cmf frame, we need to get the velocity at the correct observer frame positions

I'd appreciate your looking at this before I merge it.


source/wind.c
```diff
@@ -323,46 +333,69 @@ model_vgrad (ndom, x, v_grad)
...
   int i, j;
-  int vsub (), stuff_v ();
+  double zero_vector[3], dx[3], dx_obs[3];
+
+  zero_vector[0] = zero_vector[1] = zero_vector[2] = 0.0;
+
+
+
+
+  ds = 0.00001 * length (x);
```
saultyevil: I think this ds is in the cmf frame, right? Should we label it as ds_cmf instead?

saultyevil:
I don't see anything wrong with the changes, I just made one small comment which would make the code more self documenting.

source/wind.c
kslong: Agreed.  I will make this change.

PR #804, opened 2021-01-05
Shield

kslong:


jhmatthews:
@kslong Had a quick skim of this and looks good to me. There are some comments that have been removed that I would prefer to stay in. If they are clarifying something that is not self-evident, I think we should keep them (e.g., comments about the gaunt factor, comment about factor of ne in the kpkt rates, comments about the adiabatic cooling in kpkt routine).

kslong:
@jhmatthews  I went ahead an merged this despite your comments. .   I think if you look what I did was make a more generic comment earlier.  If, on reflection, you think something is needed, please just add something back.  Mainly though I wanted to make sure I had merged this into dev before you added the macro atom material.  (Regarding that I would suggest you change the version to 85b before merging).

jhmatthews:
Sure, that sounds fine.

Issue #423, opened 2018-08-10
Macro atom transition probabilities via Matrix solution


ssim:
As outlines in Belfast, could implement methods developed by others to avoid explicitly following internal macro atom state transitions. Based on matrix inversion method. Before attempting this would first need to scope out what's feasible:

- need for new routines to gather data into appropriate matrix (probably not expensive but need to consider how to implement)
- main question is probably about storage: matrix could get large (quantify)
- speed of inversion - need to scope out how long a call to this process would take at runtime.

jhmatthews:
This was partly implemented during the Soton 2019 code week. There is some discussion on evernote. The current code is in the hyper_macro branch and can be turned off and on via the ```#define ACCELERATE_MACRO 1``` flag in photo_gen_matom.c.

jhmatthews:
I've been testing this implementation a bit -- in which the accelerated scheme is only used in the matom emissivities part of the code -- and documenting / streamlining the code. The latter part is mostly done. The tests currently seem to produce differences with dev when using the cv_macro_benchmark pf file, beyond the statistical effect I would expect. The emissivities seem systematically lower in the new version. Possibly the new code doesn't work so well when extra ion species are included. Investigating.

jhmatthews:
This has now been "fully" implemented in the ```matom_matrix``` branch. This note is just a todo item for me to describe the code changes here.

jhmatthews:
The main changes to the code, currently in the matom_matrix branch, are as follows. This branch can be compared to dev by going to this link: https://github.com/agnwinds/python/compare/dev...matom_matrix.

**Usage:** To use the matrix scheme, select ```matrix``` as the response to the question ```Matom_transition_mode(mc_jumps,matrix)``` in the parameter file. ```mc_jumps``` uses the old method, but still uses the accelerated scheme for the emissivity calculation. This needs to documented on ReadTheDocs, and the parameter files need updating where necessary before merging into dev. 

### Main Code Changes
**Modes / infrastructure**
* new variables in the macromain structure for storing the matrix, as well as various flags.
* a new variable in the modes structure which determines whether we are storing the matrix (possibility to include a partial or adaptive option here in future). At the moment the ```macromain[n].store_matom_matrix``` directly inherits the ```modes.store_matom_matrix``` value which is set by the command line flag.
[...]

PR #806, opened 2021-01-12
Accelerated macro-atom scheme

jhmatthews:
This pull request implements a beta/development version of the accelerated macro-atom scheme described by [Ergon et al. 2018](https://ui.adsabs.harvard.edu/abs/2018A%26A...620A.156E) and also documented in a research note from Christian Vogl. The sampling of macro-atom transition probabilities is rewritten as a matrix problem as described by Christian- instead of jumps of jumping through the macro-atom machinery, we write the series of jumps as a series of matrix operations, and we do an infinite sum over each number of jumps (see document on evernote). see also #423 

The code here is only used if the ACCELERATED_MACRO flag is set to TRUE, and it currently only operates during the macro-atom emissivity calculation.

Main changes:
* A new file macro_accelerate.c which contains the relevant functions
*  Code to calculate the jumping and emission probabilities and populate the matrix
* New functions: calc_matom_matrix() and fill_kpkt_rates(). calc_matom_matrix() fills the actual matrix. fill_kpkt_rates is excised from kpkts and populates the cooling rates for kpkt destruction since this is needed by both kpkt() and the kpkt emissivity part of this calculation. We also have new functions to calculate the band limited emissivities, which make use of band-limited versions of things like the alpha_sp integrals - we have to work out what fraction of the kpkt emissivity will come out in the spectral band, for example.  
* Code to do the actual matrix inversion - for the moment using LU decomposition and then inversion.
* Code to call these functions as part of the macro atom emissivity calculation

The version is updated to 85b. Function commenting could probably be improved. Regression testing gives identical results to dev when ACCELERATED_MACRO is set to FALSE, and some small differences when the flag is on. The k-packet emissivity changes a little in a stellar model test - so currently the status is the new scheme basically works, but should perhaps be tested further, i.e. use with caution!


source/photo_gen_matom.c
```diff
@@ -161,6 +162,20 @@ get_matom_f (mode)
     commbuffer = (char *) malloc (size_of_commbuffer * sizeof (char));
 #endif
 
+    /* if we are using the accelerated macro-atom scheme then we want to allocate an array 
+       for the macro-atom probabilities and various other quantities */
+#if (ACCELERATED_MACRO == TRUE)
+    PlasmaPtr xplasma;
+    int nrows = nlevels_macro + 1;
+    double **matom_matrix = (double **) calloc (sizeof (double *), nrows);
```
saultyevil, line 170: I think there would be a memory leak here, since I can't see the pointer being free'd at the end of the function. It's probably not a problem, since we only calculate the emissivities once, so would only get calloc'd once. I think the same goes for the variable commbuffer too. Soz for the pedantic comment!

jhmatthews: Good spot - pedantry is needed on this stuff :) I'll look at it.

jhmatthews: do you know if you have to free each row of the matrix, or does free (matom_matrix) do it all?

jhmatthews: wondering whether all that is needed is:

```C
#ifdef MPI_ON 
  free (commbuffer);
#endif 
#if (ACCELERATED_MACRO == TRUE)
  free (matom_matrix);
#endif 
```

saultyevil: I was just looking into that from my old lectue notes! I think the correct thing to do is,

```c
for(i = 0; i < nrows ++i){
  free(matom_matrix[i]);
}

free(matom_matrix);
``` 

Since if you do `free(matom_matrix)` it would only free whatever that points to, which is the array of pointers. I think when those are free'd, you can't get them back to free the memory which belong to them, so you have to free each pointer/element in matom_matrix and then you can free matom_matrix.

saultyevil: I think there's the same thing in calc_matom_matrix for R_matrix and Q_matrix too.

jhmatthews: I've added a commit doing what you suggested Ed, could you maybe just check it looks right?

saultyevil: Yes! I can't see any problems with your changes and it passed the CI test!

Issue #747, opened 2020-07-29
Option to specify NWAVE to tailor spectral wavelength binning for plotting


mandyhewitt:
In order to increase the resolution of spectra, we need the ability to change the NWAVE parameter for spectral wavelength binning. It is currently defined as 10,000 in python.h. It would be good to be able to run models and change the value of NWAVE either via the command line or parameter file. For polarisation, I currently require  NWAVE of 1000 and hard code the change in my branch. We also need to change this value for Fe K-alpha work. In Tardis code this option can be tailored.

I'm not sure if this option is available already in python via running the code in a diagnostic mode nor if there is a reason it should not be adjusted but if this is a good enhancement to implement then Rahul and Mandy will try and implement it during virtual-belfast.

kslong:
I agree that for many purposes, 10,000 seems excessive.    

I would prefer that you NOT do this on the command line, but do it as a parameter and insert the fix somewhere near where this line line is set

Spectrum.type(flambda,fnu,basic)              flambda

You will want to name the variable something like

Spectrum.nwave

Note that there is an issue with this.  Currently NWAVE is a global variable and it used to allocated structures with a certain dimension. This allocation preceeds anything you read in.  So either you will need to dynamically allocate all of the spectrum arrays, or you will leave NWAVE as it is, but fill only nwave values of the array.  To do this, you will need to 

* define a new variable, nwave,  in the geo structure (probably near the other spectra related variables).  You will, I think need it in the geo structure so that you can restart models.
* replace NWAVE with geo.nwave in many places in the code.


These changes are not hard, but they will require some robust testing to make sure for example, you can re-import and continue a model.

Good luck!

kslong:
@Higginbottom   @mandyhewitt   if @kslong  were to implement this, would you wand

* to apply this just to the detailed spectra or to both.  If just to the detailed spectrum, then probably this nunber would need to be smaller than the maximum value we store.
* Assuming the answer to the above question is just the detailed spectra, then would you prefer to specify nwaves directly or delta lambda.  
* Do you have any other thoughts on what would make this feature most useful?

mandyhewitt:
@kslong I would like to be able to apply the NWAVE setting to both detailed spectra and have the possibility to set it also in the ionization cycle. 
I would prefer the value to be specified directly (as this is simple), however, I do not understand the use of  delta lambda - could you give your thoughts on this? 
[...]

PR #823, opened 2021-02-27
Add variable to control number of wavelength bins

saultyevil:
After some procrastination from writing, these are the proposed changes in Issue #747 which wanted a parameter to specify the number of wavelength bins for the spectra.

A new parameter has been added "Spectrum.nwave" which controls how many wavelength bins there are for the ionization and spectral cycle spectra. I opted for specifying the number of bins rather than delta lambda as I think this is less error prone, i.e. you may specify a delta lambda of 0.25 over a huge wavelength range by mistake and end up with a massive number of bins creating a huge spectrum file. There is minimum number of wavelength bins of 25 (defined by NWAVEMIN) but I haven't added a maximum cap.

I have updated all of the example parameter files, other than regress.old, with Spectrum.nwave 10,000 as this was the value of NWAVE prior to this change. I've also added a python script to py_progs whose purpose is to update parameter names/values--pretty similar to add_param.py.

Since these changes modified xxspec and require dynamic allocation of the spectrum arrays (and the removal of an unused array), backwards computability has, once again, unfortunately been broken. I have labelled this version as py85c and Python will exit if it does not see a certain header in the .spec_save files. 

Restarting models have been fine for me, and I've tested it on three machines to make sure it wasn't a fluke. Just be sure to do a full clean of old object files before trying to run with the new version.

The regression tests seem to indicate I didn't break anything. The difference spectra for the regression tests do look a bit odd (see plot below), I guess because the binning might be slightly different now for some reason--but the difference is (rather) small anyway. I can't see a difference in the actual overplotted spectra.

agn_macro

cv


jhmatthews:
Should this be merged @saultyevil ?

jhmatthews:
I've resolved conflicting version number using web editor (which is a nice new feature).

saultyevil:
Sorry, not just yet... after feedback from some telecon, I need to make a couple of changes to the "internal" spectra for the ionization cycles etc. Knox pointed out that it would probably be preferable for them to have a (large?) fixed value (10,000 or so) and so only the .spec spectra should be controlled by the new variable. I haven't gotten around to doing it just yet because of writing the paper etc, but it's top of my list when that's done!

saultyevil:
Huh, the Balmer test seems to be sensitive to Python version number according to the last travis build...

saultyevil:
I've fixed the number of bins for the ionization cycles to NWAVE_IONIZ = 10,000, as it was before. Of course, we could turn up this value as well as I believe someone thought this was closer to 40, 000.

With the current implementation, since it is dynamically allocated, you can specify nwave for the spectral cycles to be pretty much anything, but it can't be smaller than 10. I thought this would be better than forcing an upper limit of NWAVE_IONIZ. Doing this wouldn't guarantee backwards compatibility with older python version, since an extra variable will be required in the wind_save/spec_save to track the number of spectral cycle bins (and changing NWAVE_IONIZ, to get more bins, for a one time run will break that backwards compatibility of the spec_save anyway). I'm probably solving a problem which may not exist, but you never know :)

For the sake of code simplicity with communication between MPI processes, I opted to creating all arrays with the largest number of bins value so if the nwave for the spectrum cycles is too large, it'll also crash in the ionization cycles. This probably avoids some heart-ache, but isn't as memory efficient as it could be -- though it is not a limiting factor in terms of overall memory usage.

Below are some plots from the regression tests -- ignore the date, I haven't made changes to the code since 23 March and I'm 
only just getting around to looking at this PR again. There are _small_ differences according to the flux difference plots, but it's very hard to see the differences in the actual spectra. The differences come from _slight_ differences in binning, as I cleaned up some variables which were recalculated over and over, i.e. dlamba, instead of using pre-calculated values.

I'll be merging in tomorrow afternoon if there are no complaints.

cv_kur
agn_macro_gamma

PR #858, opened 2021-05-07
Allow GSL RNG to be continued where it left off from on model restarts

saultyevil:
By using the `--rng` flag, the GSL RNG state is saved at the end of each ion/spectral cycle. On model restart, the RNG is reloaded and also requires the `--rng` flag.

The RNG is saved in a hidden directory named `.rng_root/` and the files take the form `root_X.rng_save` where X is the process number. My tests have shown that the RNG reloading mode works, as identical "Photons contributing to the various spectra" tables and etc are obtained when using `--rng` to restart a model. When `--rng` is not used, this table is diffrent.

The regression tests, which don't use `--rng`, are unchanged by changes in this PR.


jhmatthews:
This looks great, nicely done.

PR #876, opened 2021-06-22
Changes for repeated resonances

saultyevil:
These changes do not allow the photon to interact with the same resonance twice in a row within a small volume. In these cases, the photon skips over the resonance when searching for the next interaction and can go on to interact with another resonance or continuum scatter.

Photons are also no longer repositioned after resonance scattering, so there has also been some code clean up to remove this from the photon transport.


saultyevil:
Oh and I've updated the version number to 85h.

Issue #888, opened 2021-07-13
dv/ds_max is not set in cells when dv/ds is negative


saultyevil:
There have been some changes to how we calculate dv/ds and dv/ds_max recently, to make it more consistent with our CMF treatment. When we calculate dv/ds_max in a cell, we calculate dv/ds for 10,000 photons with random directions at the corner of a cell. Then we do this:

```c
dvds_max = 0.0;
dvds_min = 1.e30;

for (n = 0; n < N_DVDS_AVE; n++)
{
  randvec (p.lmn, 1);
  dvds = dvwind_ds_cmf (&p);

  /* Find the maximum and minimum values of dvds and the direction
   * for this
   */

  if (dvds > dvds_max)
  {
    dvds_max = dvds;
    stuff_v (p.lmn, lmn);
  }
  if (dvds < dvds_min)
  {
    dvds_min = dvds;
    stuff_v (p.lmn, lmn_min);
  }
}
```

There's a problem when all 10,000 dv/ds's are negative, which is happening mostly in cells with negative divergence in my TDE model.

The negative gradients are not all that different from the positive versions, so is not a small number thing as it would fail completely otherwise. For example (for a cell which is okay),

```
Cell 130: dvds -7.680948e-05
Cell 130: dvds -2.530502e-04
Cell 130: dvds -3.440842e-04
Cell 130: dvds 1.626623e-05
Cell 130: dvds 1.220537e-05
Cell 130: dvds 9.050296e-05
```

I don't see why dv/ds can't be negative, if the flow is inflowing in those cells. We only use dv/ds_max to calculate tau_sobolev, which only cares about the absolute value, |dv/ds|. So I think if we take the absolute value of dvwind_ds_cmf we should be in the clear.

I will do this change and update the name to make it clear that it is the absolute value of dv/ds_max, unless anyone thinks we should do something else.

saultyevil:
In hindsight a more sensible thing to do would be to initialize dvds_max = -VERY_BIG so we can get negative dvds_max and still keep the sign -- because when dvds_max is eventually passed to `sobolev()` the absolute value of it is used anyway. It could be helpful in the future to keep the sign for debugging.

I have a couple of comments/questions about velocity gradients before I move on though:

[...]

Issue #750, opened 2020-08-03
kappa_ff


kslong:
kappa_ff seems to explicitly assume that ion 1 is ionized H, and ion 4 is double ionized He.  Also free emission is only possible for these two ions.  

Consider 

- Checking that the ions are the ones that are expected

- Consider adding more highly ionized species into the FF calculation

Higginbottom:
If the free free gaunt factor data file is not present, then it is true that 

a) the ions are hard wired and
b) only H and He are used.

I think that the second point is because the error in not knowing the gaunt factor is probably greater than including minority ions. It was written this way to preserve the old behaviour when the gaunt factor file was included.

However, I believe that if the gaunt factor data is read in - then al ions contribute and there is no assumption.

We may need to fix the fallback position, or do away with it entirely and error if the gaunt factor file is not present.

Nick

kslong:
This really was a misreading of the code. Some documentation has been added to try to avoid this question in future.

The basic story is as follows.  

* In the original version of the code, it is true that the ff opacity was estimated based only on H and He, and that estimate of the ff opacity included no gaunt factor
*At some point, he code was modified to include all ions and to include a gaunt factor based on a table provided by Sutherland, R.~S. 1998, MNRAS, 300, 321. What happens in this case, is that a prefactor to the ff opacity is calculated at the for each ionization cycle and stored in  plasmamain[].kappa_ff_factor
* Today, unless one fails to read in the Sutherland table in the atomic data, the gaunt factors are calculated on this basis.  However if one leaves the Sutherland data out of the atomic data, the code will go back to the original approach.

In inspecting the code, it seems likely this code could be made a bit more transparent.  Most of the routines are in radiation.c, but the calculation of gaunt_ff based on Sutherland is in emission.c

Having done this, I am closing his issue.

PR #899, opened 2021-08-18
Bug888 dvds max

kslong:
This should assure that the calculation of dvds_wind_max is always positive, which address #888 
Also, documentation was added to address the confusion ksl had about kappa_ff and specifically the gaunt factor #750


saultyevil:
Using the absolute value looks good to me, which is also what I settled on doing originally.

PR #945, opened 2022-04-01
Changes to inspect wind to report line luminosities

jhmatthews:
This adds functionality to inspect wind to print out all line luminosities that appear in the requested spectrum wavelength range. It does this by sampling the (correctly normalsied ) emission probability for each bound-bound downwards transition and multiplying this by the upper level emissivity. 

The code creates a new folder called matom_linelum_root which contains the line luminosities from each upper level in macro-atoms. The files look like this:
```
# model name disk7_big
# Line luminosities from upper level 3 from 3000.00 to 24000.00 Angstroms
# Lower Levels:                0           1           2
# Line Wavelengths:       972.27      4861.35     18750.80
       nwind            x            z    i    j       inwind     wind_vol   plasma_vol  LowerLev000    LowerLev001    LowerLev002
           0   2.7772e+10   3.7500e+06    0    0           -2   0.0000e+00    0.0000e+00    0.0000e+00    0.0000e+00    0.0000e+00
```
where the header info tells you which lines correspond to which lower levels. The lower levels are the config numbers as stored in the config structures. the columns have a similar format to windsave2table or py_wind outputs. There's also a file line_map.txt which contains info on which levels each lines map to:
```
# model name disk7_big
# which line wavelengths (Angstroms) correspond to which upper and lower levels
upper lower wavelength
1 0      1215.34
2 0      1025.44
2 1      6562.84
3 0       972.27
```
I've written a python script to read these and plot up which I'll circulate.

### DIFF
--- a/source/inspect_wind.c
+++ b/source/inspect_wind.c
@@ -29,8 +29,12 @@
 //OLD #include "import.h"
 
 
-char inroot[LINELENGTH], outroot[LINELENGTH], model_file[LINELENGTH];
+char inroot[LINELENGTH], outroot[LINELENGTH], model_file[LINELENGTH], folder[LINELENGTH];
 int model_flag, ksl_flag, cmf2obs_flag, obs2cmf_flag;
+
+double line_matom_lum_single (double lum[], PlasmaPtr xplasma, int uplvl);
+int line_matom_lum (int uplvl);
+int create_matom_level_map ();
 
 /**********************************************************/
 /**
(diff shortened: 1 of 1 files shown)


jhmatthews:
I think there's a mistake with higher ions that H1 at the moment (e.g. He II goes from 77 to 2 in the line mapping which must be wrong).

jhmatthews:
I've fixed the mistake in this code (I was printing out the number of the bbd_jump rather than the config to line_map.txt). Should work now and can probably be merged. 

Allows one to produce plots like this:
line_lums_disk7_big
:

PR #947, opened 2022-04-06
Docs reorganisation and cleanup with some additions

jhmatthews:
This is a proposed documentation reorganisation. The basic changes are:

* consolidate and reorganise folders into a more sensible hierarchical structure
* remove duplication of e.g. convergence info
* add more python modules to the autodoc entry 
* make a few of these docstrings more sphinx/rst compliant
* some additional docs on the macro-atom emissivity calculation including a note about what the units are 
* a note in the python header file about what these emissivies are

The docs structure that would produced by this commit can be previewed [here](https://people.ast.cam.ac.uk/~matthews/html_docs_proposal)


jhmatthews:
@kslong @lazygun37 @Higginbottom if you're happy with this structure I'll merge it but let me know if you have any suggested changes. Knox this also addresses your comment about the emissivity units.

kslong:
Yes,  I think this is an improvement.  Merge when ready, Gridley.

PR #950, opened 2022-05-09
Upstream dev

Higginbottom:
Made a further correction to rad_hydro_files and also added some new output to run-c to show the origin of luminosity escaping the simulation


kslong:
@Higginbottom I also don't understand why this is causing things to "re-merge".  from 9 months ago.  I tried looking for the commits in my version of dev, and I could get to some but not all of them.

I assume you have a version of Python on your machine that is not a fork.  If not I would set one up, probably.   Create and push a test branch to git hub and try to squash and merge into that one before directly merging in dev.  Maybe @jhmatthews understands this all better.

Higginbottom:
My version of python is a fork from the main repo. And before I submitted the pull request, I did a pull from the main dev branch - almost nothing changed, as expected because I keep it more or less up to date. Confusing!

jhmatthews:
Yeah, I don't fully understand the reason for this. It seems a bit subtle. Found this on stack overflow: https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch

I don't really understand the main issue - there's a few comments along the lines of "the root cause is that a PR is a diff of the source branch tip against the common ancestor of the source branch and the target branch. It will therefore show all changes on the source branch up to the common ancestor and will not take into account any changes that may have occurred on the target branch." But I can't fully parse that or see why it would happen if you pull from upstream regularly. 

I don't remember this ever happening to me, so it's possible it depends a bit on the local user's merge strategy or git version.  also wonder if it's something to do with having used git revert at some point?  

So, although it's a bit messy, I'd not worry too much if there are some additional commits in the commit history **as long as the actual file diff is OK**.  

**BUT** the file diff is not OK here - there's a conflict that would stop the code compiling in estimators_simple, and there seems to be a bunch of atomic and model data removed from ```xdata/``` - so these need fixing, right?

Higginbottom:
Oh man - I didn't even notice the data stuff. Why on earth has that happened. I guess it's something to do with me downloading the other data repo to do the stellar atmospheres experiment. Too hard.....

Higginbottom:
OK, so I think I've fixed up the atomic data - no idea *what* was going on there, xdata seemed to have become nested inside  subdirectories... The diff now just shows what I think should change ....

kslong:
Great

From: Higginbottom ***@***.***>
Reply-To: agnwinds/python ***@***.***>
Date: Tuesday, May 10, 2022 at 4:35 AM
To: agnwinds/python ***@***.***>
Cc: Knox Long ***@***.***>, Review requested ***@***.***>
Subject: Re: [agnwinds/python] Upstream dev (PR #950)


OK, so I think I've fixed up the atomic data - no idea what was going on there, xdata seemed to have become nested inside subdirectories... The diff now just shows what I think should change ....

kslong:
I don’t understand this either but I suggest you squash as you merge so it Does not have the possibility of creating a crazy set of logs

Sent from my iPhone

On May 10, 2022, at 3:38 AM, Higginbottom ***@***.***> wrote:

﻿

My version of python is a fork from the main repo. And before I submitted the pull request, I did a pull from the main dev branch - almost nothing changed, as expected because I keep it more or less up to date. Confusing!

PR #991, opened 2023-01-09
Implementing the Auger process in macro-atoms

jhmatthews:
This pull request includes updates made by myself and Stuart in Summer 2022 to improve our treatment of the physical processes associated with Fe fluorescence and Auger ionisation. 

A more detailed description of these changes can be found on evernote and in the attached PDF.

To implement this in python, we made the following changes:
* added code to matom.c and macro_accelerate.c to simulate the Auger process. These are analogous pieces of code but one uses MC jumps and one uses the matrix scheme.
* Added a new structure called auger_macro which stores the Auger atomic data for macro-atoms
* Wrote code in atomicdata.c and atomicdata_init.c to initialise these structures and populate them with atomic data we read in.
    *Create new (experimental!) atomic datasets to test our process including:
    * A line list with radiative transitions. We removed some lines with 0 frequency / infinite wavelength caused by having resolution to only 7 (?) sig fig in energy levels. These should be long wavelength lines anyway or in some cases the levels may be fully degenerate to the accuracy of Chianti.
    * These datasets came from Stuart’s code, the ultimate data source is probably Chianti. 
    * See more details below
* Increased a number of variables like:
    * NLEVELS_MACRO from 200 to 500
    * NTOP_PHOT from 400 too 1000
    * NBB_JUMPS from 100 to 200
    * NBF_JUMPS from 100 to 200

Stuart and I have inspected the new code and believe it should be safe to merge in -- in the sense that it won't break or alter the behaviour of the code when used with other "normal" atomic datasets. 

The two question here to resolve **before we merge** are:
* at this stage we haven't run any "standard" models to check they don't change. To do.
* Should we change the define statements like NLEVELS_MACRO back to their old versions to avoid increasing memory usage in models without auger? Thoughts @kslong?

Once these questions are resolved we can probably merge this. 

[Belfast22-auger_macro.pdf](https://github.com/agnwinds/python/files/10373605/Belfast22-auger_macro.pdf)


jhmatthews:
For the moment I've reduced NLEVELS_MACRO,NTOP_PHOT, NBB_JUMPS and NBF_JUMPS back to their old values but but a note next to each of these beginning with "AUGER NOTE"

PR #1006, opened 2023-07-20
Change from cartesian to spherical fluxes

nscepi:
I changed the fluxes that are computed in estimators_simple.c so that those are stored in spherical coordinates instead of cartesian coordinates. This avoids changing frame again in PLUTO and minimize errors due to a coarse grid. I changed the names of the structures like F_UV_x/y/z to F_UV_theta/phi/r.


kslong:
This should not be merged into main.  We need to merge into dev instead

PR #1009, opened 2023-08-17
Grid save: allow user to save wind grid & quit without running cycles

jhmatthews:
This proposed change adds a new flag, ```--grid-only``` to be used from the CL. If this flag is on, it sets ```modes.quit_after_wind_defined = 1```. 

If this mode is on, the code quits before running any cycles or producing any spectra, but will save a wind_save file. Much of that wind save file will be useless, but it will allow a user to check the wind grid and the populated densities and velocities, which could be useful for verifying imported models, or for seeing how different wind parameters change the density-velocity space populated by the model.

### DIFF
--- a/source/parse.c
+++ b/source/parse.c
@@ -253,6 +253,12 @@ parse_command_line (argc, argv)
         j = i;
       }
 
+      else if (strcmp (argv[i], "--grid-only") == 0)
+      {
+        modes.quit_after_wind_defined = 1;
+        j = i;
+      }
+
       else if (strcmp (argv[i], "--version") == 0)
       {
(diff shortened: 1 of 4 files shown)


jhmatthews:
As an aside, I realised I don't really know what I'm supposed to use when exiting the code. i.e. should it be Exit() or exit(), and what code should I use? This is not consistent in the code, so in this case I just copied the other exit(0) commands in python.c, but I suspect these are all using the wrong exit().

jhmatthews:
Merging this as it seems to work OK for me. If anyone has problems with it let me know.

Issue #1008, opened 2023-08-15
Add the colour correction


jhmatthews:
Scattering/radiative transfer in the disc atmosphere can change the radiation temperature of the emitted radiation from an accretion disc - in particular it often makes the shape of the spectrum hotter even if the effective temperature is the same and set by local energy dissipation. This effect is often parameterised via a colour correction $f$ such that  

$$B_\nu (\nu, T) \to \frac{1}{f^4} B_\nu (\nu, fT)$$

[ooh, github has math mode now]

where $f$ is in general a function of at least temperature if not also surface density and other things. 

We should implement this colour correction factor as an option in python, which can be done fairly straightforwardly. There are various options. [Done et al. 2012](https://ui.adsabs.harvard.edu/abs/2012MNRAS.420.1848D/abstract) give a simple form of the colour correction in which $f=1$ for $T<3\times10^4~{\rm K}$, and for $T>3\times10^4~{\rm K}$

$f=\left(\frac{T}{3\times10^4~{\rm K}} \right)^{0.82}$

I suggest we implement this as a starting point. 

Proposed implementation:
* either a new disc temperature mode called "standard_fcol" or a separate question called "colour_correction". Alternatively, this could be part of "disk radiation type" which could also have bb_fcol. These questions would turn on a new mode stored in geo, presumably. 
* new function that returns the colour correction when queried with a temperature. This function could check the mode variable.  
* modify the relevant fluxes in the disc calculation, and also the temperature. Probably safest to only apply this for BB at the moment.

lazygun37:
I actually have an implementation (in python -- the language) of the colour correction factor described by https://arxiv.org/abs/1809.05134

It's a little more complicated, for various reasons, but probably would still be kind of easy to port to Python (the code).

jhmatthews:
That might be handy. I think I'll start by adding the done colour correction then we can always add more (I'll write it in such a way that we can add different colour correction prescriptions easily. 

I realised two things. 
* it makes more sense to define this as an additional spectral type rather than in the temperature profile question. You should still have the freedom to apply a colour correction to any temperature profile
* I think I'm write that each annulus will still emit exactly the same luminosity, so I think this actually only matters when choosing the photon frequencies. This is neat because it means that all the disk initialisation and luminosity calculations are identical, it's only the photon generation that needs a factor of "f" in the temperature when it calls the routine that generates a BB distribution of photons. Almost like we planned it all along.

lazygun37:
[...]

PR #1017, opened 2023-09-14
Implementing a colour correction

jhmatthews:
This is a first implementation of the colour correction for discs in python (see #1008). The Colour correction can be used through the following options:

```
Disk.rad_type_to_make_wind(bb,models,fcol) fcol
Disk.rad_type_in_final_spectrum(bb,models,uniform,mono,fcol) fcol
```

If you select fcol in the first one, then another question will be asked: 

```
Disk.colour_correction(done) done
```

for which the only current option is "done".


jhmatthews:
Can I merge this @kslong ?

kslong:
@jhmatthews  yes, I think you can merge it, though I think the name fcol is not very user friendly, as what one means is a modified bb, in that since bb_fcol, as you described in the issue might have been clearer, or even color_corrected_bb.  

We have used more descriptive terms in other places, and I would prefer that.  As it's a minimum match one does not have to type everything.  Of course, for this reason one would want to check that using bb_corrected, or something would not mess up simple bb models.  Maybe mod_bb.  "done' is equally confusing, I would put Done in the command, though in fact, I think we do a lower case transition.  ...

OK here is what I would suggests

````
Disk.rad_type_to_make_wind(bb, mod_bb, models) mod_bb
Disk.rad_type_in_final_spectrum(bb, mod_bb, models,uniform,mono) mod_bb

Disk.bb_prescription(Done) Done
````

with the latter suggestion in part to avoid confusion between color and colour.

jhmatthews:
Will make done capitalised. I could call it Done2012 or something to make it more explicit. 

mod_bb I'm less in favour of since this does explicitly set a colour correction at the moment and modified BB has all sorts of meanings - f or fcol is the general notation used in the literature. I don't make making it more explicit though.

jhmatthews:
I do think also there's an argument for brevity in the parameter file(i.e. colour_corrected_bb would be too long) particularly if we are likely to add more options and keep the (options,) format for the keywords. 

For documenting the parameter options, can I just manually update the readthedocs page?

kslong:
I liked mod_bb = modified_bb; I think fcol is an "inside" term for the AGN community, especially as you are planning additional versions of Done, or Done12.  I am not too bothered by longish names, as the switch from numbers to names was intended to make things more self evident.  But you should decide, make whatever changes you want, and go ahead and merge.

jhmatthews:
I've now modified the syntax/keywords as Knox suggests, so this can be merged once we've sorted the unit tests.

jhmatthews:
Merging this as tests now pass with reverted change in dev. Docs probably need improving a bit but I'm struggling to build them locally at the moment

PR #1027, opened 2023-09-28
HPC RSE: Matrix Acceleration and Unit Testing Framework

Edward-RSE:
This PR will merge in the changes made for the HPC RSE project, relating to matrix parallelisation using CUDA and the implementation of a unit testing framework.

So in this PR, we have:

- The naïve implementation of using cuSolver to solve _square_* linear systems and calculate the inverse of _square_ matrices. There are two new wrapper functions for this: `solve_matrix` and `invert_matrix`. 
- Conditional compilation between CUDA and GSL. To enable CUDA, use the switch `--with-cuda` with the configure script.
- A unit testing framework, using [Cunity](https://cunity.gitlab.io/cunit/). This was part of the project to avoid any regression being introduced as things were shuffled around and re-implemented in cuSolver. So far there are two test suites: one to test the GSL and cuSolver matrix code, and another example suite testing some of the Compton scattering functions.
- Cookbook level documentation on: 1) How to write unit tests and use the unit tests in the Cunity framework. 2) Usage, compilation and implementation documentation for the cuSolver code.
- Changes to the GitHub action to run (and potentially fail the action**) the unit tests

Apologies for the messy commit history and large number of files in the commit (most of which are CUnity). I had intended to go back and squash a number of commits together, but some commits had multiple parents which stopped me from being able to clean as far back as I would have liked. It's still possible to squash this feature branch into dev instead, if that's preferred. 

*Both the GSL and cuSolver implementations are set up for square matrices. It should, however, be easy to extend both implementations to arbitrary shaped matrices.
**That is fail when a unit test fails.


jhmatthews:
I guess we can probably merge this @Edward-RSE ? Are the tests in the github workflow? I couldn't fully tell from the yml file but couldn't see the test output in the log.

Edward-RSE:
I'm happy for this to be merged. 

The unit tests run in the "build" part of the workflow, at the bottom of the "make gh_workflow_install" task. I think this [link](https://github.com/agnwinds/python/actions/runs/6393339269/job/17352371682?pr=1027#step:4:4826) may take you to the right line in the logs.

It will be easy enough to move it elsewhere in the workflow if that's preferable.

PR #1028, opened 2023-09-28
Tau

kslong:
This change, suggested largely by James, fixes the difference between Python and Sedona for some well-defined cases.


jhmatthews:
Merging this since we know it works now.

Issue #1026, opened 2023-09-27
ds_to_plane inconsistent with other ds routines: results in errors in imported models


jhmatthews:
In issue #1025 I've fixed one problem, but I'm now finding that some photons launched from underneath the wind don't pass through it in an imported model. Look at the purple photon here:

![traject_fix1](https://github.com/agnwinds/python/assets/3007249/b8e7fedc-47e5-45da-8f77-6226d6d67027)

For imported models, what should happen in this case is that on the first call to translate in space, ds_to_wind should calculate the following distances:
* a) the distance to the edge of the domain defined by rmax
* b) the distance to a plane at the lowest z of the wind defined by zdom[ndom].windplane[0]
* c) the distance to a plane at the highest z the wind defined by zdom[ndom].windplane[1]
* d) the distance to cylinders at rho_min and rho_max.

It should then find whichever of these distances is smallest and return that value. In this case, b) should be the smallest distance, but the code instead finds a). Why does this happen? 

First note that the initial photon position is: 8.2082e+09 4.8195e+10 -1.0000e-06 
Secondly note this means the photon is going "downwards" i.e. is emerging from below the disk plane - that's why the z position is negative EPSILON (see line 173 of disk_photon_gen.c). 
Thirdly note that at the start of ds_to_wind we initially set ds to the distance to the edge of the domain and the following bit of code is just for imported models. 

Here's the relevant bit of code in ds_to_wind:
```C
      x = ds_to_plane (&zdom[ndom].windplane[0], &ptest);
      if (x > 0 && x < ds)
      {
        stuff_phot (pp, &qtest);
        move_phot (&qtest, x);
        rho = sqrt (qtest.x[0] * qtest.x[0] + qtest.x[1] * qtest.x[1]);
        if (zdom[ndom].wind_rhomin_at_disk <= rho && rho <= zdom[ndom].wind_rhomax_at_disk)
        {
          ds = x;
          *ndom_current = ndom;
          xxxbound = BOUND_ZMIN;
        }
      }
      x = ds_to_plane (&zdom[ndom].windplane[1], &ptest);
      if (x > 0 && x < ds)
      {
        stuff_phot (pp, &qtest);
        move_phot (&qtest, x);
        rho = sqrt (qtest.x[0] * qtest.x[0] + qtest.x[1] * qtest.x[1]);
        if (zdom[ndom].wind_rhomin_at_disk <= rho && rho <= zdom[ndom].wind_rhomax_at_disk)
        {
[...]

PR #1029, opened 2023-10-02
attempt to fix translation issues in imported models

### LINKED ISSUE #1025: Code doesn't deal properly with photons that enter then exit wind in imported models
I'd been seeing strange patterns in photon numbers in an imported JEDSAD model (model files on slack in the mhd-winds channel. The basic issue is that there are big streaks in the photon number:
ntot (1)
)
I noticed that if you plot the photon positions each call in translate, you get the following plot for a random subset:

traject

This plot shows the locations of the photons just before the call to translate is made. The directions are fine, but some subset of photons passes through the wind correctly (i.e. has lots of points corresponding to passing through each cell) - another subset seems to get to the edge of the wind then bypasses the wind entirely.

I tracked in detail what happens to one photon and this is the result (annotated plot with points from actual code output for photon 30,000: 

Screenshot 2023-09-26 at 12 03 27

The basic problem is that, if the photon exits the wind after entering it, the code has no way of allowing it to enter again - because the starting ds it calculates on the call to translate in space after existing the wind is already taking it through the entire wind. 

I think this particular problem may only appear in imported models where we don't define wind cones, but I'm not certain. It also begs the question whether.

The fix will clearly have to catch this case and increment ds from some small number to check if you hit the wind, but it's a little confusing to think how best to do this.

jhmatthews:
This PR aims to fix #1025 and #1026 

The first change -- for #1025 -- is a bit complex to understand and is made in translate_in_space() and only in imported models. Originally, for imported models, in translate in space we were first calculating the distance to the edge of a box surrounding the wind, and then move it there. Then, when we are inside this box, we do the same again, but now this results in moving the photon to the edge of the box. The crux of this is that photons which "stayed in the wind" were fine, but photons which entered then exit were incorrectly transported to the edge of the box. My proposed fix is to manually increment ds cell by cell in the case that the photon position is inside rhomin and rhomax of the wind. 

The second change involves ds_to_plane - the windplane that we use to check is only defined in the northern hemisphere. So rather than doing an additional check, I added something that reflects a test version of the photon ptr for photons that are moving downwards in the southern hemisphere. @kslong you might want to check this seems reasonable, and also I have no idea if this will break the other places we use ds_to_plane. 

Probably needs some commenting and could perhaps do with a unit test or two added to test ds routines. I could perhaps add this once Ed's changes are merged.


jhmatthews:
I'm not sure that these fixes will catch every single problem case, but I have checked that when you "fudge" translate in space by just always translating by 1e7, you get the same results as for this fix. So I think it works at least for the JEDSAD model. here is before and after with this fix. Before:
ntot_onecycle_dev master txt


After: 
ntot_onecycle_fix_ds master txt


Note there are still a few streaky looking patterns, but here is the "fudge" case where ds is always 1e7 in translate in space:
ntot_onecycle_fudge master txt

jhmatthews:
This is basically working now, but there are probably two things to check:
a) does this affect performance much, probably only relevant to imported models 
b) this currently fails a test I wrote in which the photon is moved 50% along a path and then has its direction reversed. It means I'm not totally sure how ds_to_plane is supposed to work in this instance, and it still might be working incorrectly for photons that scatter back across a midplane, for example.

jhmatthews:
James and Knox discussed this at Soton24. They agreed that the old behaviour of ds_to_plane should, in general, be preserved to avoid breaking other bits of code, but there should be a mode variable added, called something like force_positive_z. Thus, in photon2d.c the call would become 

```C
ds_to_plane (&plane_m2_far, p, TRUE);
```

and elsewhere the call would be 

```C
ds_to_plane (&plane_m2_far, p, FALSe);
```

The ds_to_plane function would become

```C
double
ds_to_plane (pl, p, force_positive_z)
     struct plane *pl;
     struct photon *p;
     int force_positive_z;
{
  double denom, diff[3], numer;
  struct photon ptest;

  stuff_phot(p, &ptest);
  if (ptest.x[2] < 0 && ptest.lmn[2] < 0 && (force_positive_z == TRUE))
  {
    ptest.x[2] = -ptest.x[2]; /* force the photon to be in the positive x,z quadrant */
    ptest.lmn[2] = -ptest.lmn[2]; /* force the photon to moving in the positive z direction */
  }

  if ((denom = dot (ptest.lmn, pl->lmn)) == 0)
    return (VERY_BIG);

  vsub (pl->x, ptest.x, diff);

  numer = dot (diff, pl->lmn);

  return (numer / denom);
}
```

jhmatthews:
As well as the above, I need to add the translate test to the repo and check it passes.

jhmatthews:
I've implemented the above change, which should preserve the old behaviour of ds_to_plane in all cases except when using TRUE in the new mode argument, which is what is used when you want to have a symmetric plane above and below the disk. 
This also allows me sing the praises of the unit testing framework which helped me catch a small error in what I'd done initially, thanks @Edward-RSE - I've fixed these tests and they now pass as I expect.

jhmatthews:
I've checked also that this PR produces smooth photon numbers in the imported JEDSAD model. It produces absolutely  identical outputs for the CV standard model as desired. Merging.

Issue #1045, opened 2024-01-09
Macro-atom speedup changes are cause incorrect ionization states


kslong:
There are several indications that in one of the recent updates to Python intended to speed up macro atom calculations something has failed.  There are several threads of issues, not the least of which when Macro atoms are invoked cloudy-python comparisons are failing for the macro atoms, which was not the case in 2140b19c from 231029.  The problem is apparent in 3a49b22efc, which also failed James' Balmer test.

kslong:
I found a mistake caused the cloudy/python comparisons to fail entirely.  The results are still not as good as before, but rather than a programming error this might be that the integrations are not as acurate.

This is what py87f produced (before the change)

<img width="748" alt="image" src="https://github.com/agnwinds/python/assets/2530901/5fa42f10-ab42-4663-849a-f3c1b832ae0c">

and this is what 87g produces after the fix

<img width="767" alt="image" src="https://github.com/agnwinds/python/assets/2530901/b59abc91-8af3-46f0-b816-b9b93542b29f">

As one can see, both 87f and 87g are failing at very low IP, but the problem is more "long lasting" for the new cruder, but faster integration routine.

jhmatthews:
For reference, here is the results a similar test I got for my thesis: 
![Screenshot 2024-01-11 at 11 39 41](https://github.com/agnwinds/python/assets/3007249/2e16bca8-7098-49d8-a5be-45adcc29b229)

@kslong obviously the low IP needs sorting out, but in some sense the comparison before the change was already an improvement in some areas on my calculations. Is this using your new atomic dataset for He?

kslong:
No, I have been testing with data/h10_hetop_standard80.dat

The results with the branch indicated above from 231019 are very close to what is in your thesis @jhmatthews   The differences are due to the fact that my Python grid did not go all the way to -8 in IP.  The model comparisons I posted above are for converged models, which I also believe is what your thesis likely contains.  

I've now looked at how the ionization changes with ionization cycle in one of the models where there is a difference between 87f and 87h.  In the first cycle, there is good agreement (with the fix I made yesterday), and but they arrive a different final ionization fractions after 5 or so cycles, specifically 87f has significantly more He I than 87h.  

Unfortunately that fact that the problem does not express itself in the initial cycle, it means that tracking down the problem is going to be more difficult

kslong:
[...]

PR #1046, opened 2024-01-10
Attempted correction to Fe macro data

ssim:
Attempted to correct the photoinization cross sections for the Fe macro atom data. Just want to replace the previous Z26-I17-27_phixs.dat file with a new one...


jhmatthews:
Need me to check anything Stuart, or can we just merge this?

ssim:
I think it should be fine to merge - I wouldn't promise that anything is right...but it should be better than last time :)

jhmatthews:
Yep, I will rerun the AGN model I ran before to see what changes. This will fail the tests because of issue #1045 , I think, but I'm merging since no code changes and that's a separate issue.

Issue #295, opened 2017-09-05
Compton heating - energy leak


Higginbottom:
When we originally implemented Compton heating, we did it via a continuum opacity which attenuates passing photon packets and transfers energy to the cell it is passing through. This is logged as heat_comp. This approach is based upon the idea of averaging the Compton heating effect over electron velocities and scattering angles. 
This method changed slightly in late 2015 when we tried to implement a method that would give rise to a Compton hump (at least in part) - i.e. we wanted to properly change the frequency of photons scattering off electrons. Now, we still work out the heating effect via the opacity method, but we do not attenuate photon packets at that point. Instead, when a photon scatters via electron scattering, we work out a new direction for the photon, and use this to compute an energy change. We use this to reduce the photon frequency, and the photon weight is reduced by the same amount.

This seems to work correctly however there is a disconnect between the energy added to the cell, and lost from photons. In the most extreme case, in optically thin cells, there might be no interactions at all, so energy is added to the cell via the opacity but not lost from the photons. 

Compton heating is dominant in low density gas exposed to an intense radiation field. Just because the radiation field is intense doesn't necessarily mean lots of photon packets though, each of a few packets might have huge weight. This would mean that if we just went over to only heating via scattering events we would not reach the Compton temperature.

So, we have three choices:

1: Keep the code as is - this introduces a (usually small) energy leak where photons don't lose weight even though the cell is being heated

2: Go over entirely to a scattering model - so we only transfer energy during scattering events. Unless we come up with something very clever this will result in catastrophic problems in models where we expect to reach the Compton temperature.

3: Go back to attenuating the photon packets via opacity. If we then do *not* attenuate the packets when scattering, we will violate photon numbers. We take a packet of weight w, frequency nu, and hence n=w/hnu then reduce nu and increase n. How bad is this exactly??

Higginbottom:
I've done a small number of tests, and it looks like if you make cells optically thick, then the Compton heating from opacity and the energy loss from photons during Compton scattering become similar. So it looks like the method we do at the moment is about right. Its just that we do not assure that the energy balance is maintained.

kslong:
It seems to me we should stick with our existing model for now.  

[...]

PR #1050, opened 2024-02-15
Don't record Compton heating as a contributor to kpkt_abs

jhmatthews:
This is a (hopefully temporary) fix for the problems we have been having with TDE models where the wind (or k->r) luminosity in spectral cycles is much higher than it should be (or compared to the ion cycles). 

The issue is that we include Compton and induced Compton heating as a contributor to ```kpkt_abs```, although we don't actually allow r->k transitions from Compton, or k->r transitions. This means the ionization cycles were sort of doing things "correctly" in that they aren't modelling the flow of energy through the thermal pool from Compton processes (correctly in the case that they are dominant and roughly balance each other, or are both negligible). However, in the spectral cycles we have all this extra energy in kpkt_abs which has to come out somewhere, so it comes out in all the free-free and bound-free processes and so on. 

Note, that we do reweight packets when Compton scatters happen, the thing that is missing is the machinery to allow Compton processes to communcate correctly/self-consistently with the k-packet pool.  

This current fix simply moves a line of code up so that kpkt_abs is not incremented by Compton heating. This should be considered a "bandaid" solution and in the near future we should improve this. 

See also #53 and #295

### DIFF
--- a/source/estimators_macro.c
+++ b/source/estimators_macro.c
@@ -223,7 +223,13 @@ bf_estimators_increment (one, p, ds)
 
   y = weight_of_packet * kappa_ff (xplasma, freq_av) * ds;
 
-  xplasma->heat_ff += heat_contribution = y;    // record ff hea        
+  xplasma->heat_ff += heat_contribution = y;    // record ff hea   
+
+  /* This heat contribution is also the contibution to making k-packets in this volume. So we record it. */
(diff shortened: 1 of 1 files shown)


kslong:
Please create an issue for this.  While this may be OK for the moment, it needs to be high on our priority list to fix.

jhmatthews:
Yep, on my todo list.

jhmatthews:
Merging this, but we should sort this out properly as discussed in #1053

PR #1056, opened 2024-03-11
Version change to py88

kslong:
This simply closes out py87 all versions, with change in version to py88


kslong:
This merge encapsulates all of the changes that occurred in the various versions of py87.

PR #1078, opened 2024-05-24
Update communicate_plasma.c

jhmatthews:
Edit to length of communication buffer (NFORCE_DIRECTIONS) rather than (NUM_RAD_FORCE) in doubles. Also nions in ints.

### DIFF
--- a/source/communicate_plasma.c
+++ b/source/communicate_plasma.c
@@ -27,7 +27,7 @@
  *
  * @details
  *
- * The communication pattern is as outlined in broadcast_updated_plasma_properties.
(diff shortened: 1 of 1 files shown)


jhmatthews:
It looks to me like NFORCE_DIRECTIONS is the right one to use here (because it's the length of the variables being communicated - NUM_RAD_FORCE_DIRECTIONS is for dmo_dt instead. Although there are clearly issues here where other things have hardwired [4] allocations rather than being linked to NFORCE_DIRECTIONS, and potential for mistakes in calculating the buffer size.

jhmatthews:
I will merge this, but may fix this up slightly afterwards.

PR #1080, opened 2024-05-28
Update matom.c

ssim:
A fix for choosing the auger process (think was just a code error before?)


jhmatthews:
I agree this is a correct fix, merging.

PR #1081, opened 2024-05-30
Fix pack definitions to use define statements and make consistent 

jhmatthews:
… fluxes and rad forces


jhmatthews:
This PR was nonsense, it completely broke the code and CI didn't spot it - because we only compile in CI with gcc and so MPI_ON is set to FALSE. Firstly, woops, but also we should try to change the workflow to catch anything broken only in parallel. This was literally a deleted line that means the code can't compile with mpicc!

PR #1091, opened 2024-08-06
macro_xs: include more xsections for the simple ions in macro-atom mode + 3 other fixes

jhmatthews:
This PR makes four changes:
* includes more simple-atom cross-sections when calculating the macro-atom b-f opacity and heating. This results in more accurate b-f heating rates when metal b-f heating is important 
* minor typo fix in py_plot_util 
* only run setup_py_dir when you are rank 0
* Allow the code to use ion fractions not estimated from full macro-atom estimators. 

This should not be merged yet, I'd like to test the first change a bit more and also allow the user to access the bottom mode properly.


jhmatthews:
This can be merged now. This fix produces much better heating and cooling matches with Cloudy in the intermediate U regime: [hc_macro_xs.pdf](https://github.com/user-attachments/files/16561413/hc_macro_xs.pdf) 

Compared to the old behaviour here: 
[hc_dev_macro_problems.pdf](https://github.com/user-attachments/files/16561470/hc_dev_macro_problems.pdf)


Regression tests produce relatively small changes (only in the macro-atom runs), e.g.: 
agn_macro


Please note this increases runtime in some ionization cycles, which is fairly unavoidable. In agn_macro regression test, the photon transport takes 16 seconds compared to 12 before.

jhmatthews:
Final note: this includes a mode for not using macro_pops for ionization fractions. I still need to add a way for the user to actually use this.

PR #1094, opened 2024-08-15
Proposed cleanup of rdpar choices logic to make radiation sources I/O work 

jhmatthews:


source/rdpar_init.c
```diff
@@ -233,19 +231,13 @@ get_choices (question, choices, qstruct)
...
   {
-    for (j = 0; j < num_choices; j++)
+    for (j = 0; j < qstruct->n; j++)
     {
       if (strncmp (cur_choices[i], qstruct->choices[j], strlen (qstruct->choices[j])) == 0)
       {
         cur_values[i] = qstruct->vals[j];
         break;
       }
-      if (j == num_choices)
```
jhmatthews, line 243: Removed this because it can never happen inside this for loop

source/rdpar_init.c
```diff
@@ -157,11 +161,8 @@ get_choices (question, choices, qstruct)
      char *choices;
      struct rdpar_choices *qstruct;
 {
-  int num_choices = 8;
```
jhmatthews, line 160: This was the offending line

source/python.h
```diff
@@ -1734,11 +1733,12 @@ extern int xxxbound;
  * for each input variable.  At present, this is only
  * used for the selection of spec_types
  */
+#define MAX_RDPAR_CHOICES 20
```
jhmatthews, line 1736: Increased this and established link with structures to prevent e.g. segfaults.

PR #1101, opened 2024-10-08
Proposed minimal banding change to prevent fewer than 7 ionization bands

jhmatthews:
This pull request (PR #1101) introduces updates to the banding strategy and other minor changes. Specifically: 

* Enforcing a minimum of seven bands for ionization, otherwise defaulting to the old 10 hardwired bands we used to use
* New function ion_bands_init to replace freqs_init and ensure the ionization banding is carried out in a separate function
* Small adjustments to flags and comments to clarify the behavior of the bands in specific cases.
* New Function: check_appropriate_banding provides some rudimentary checks on banding, but could be expanded

Testing ongoing.

### DIFF
--- a/source/bands.c
+++ b/source/bands.c
@@ -716,6 +716,9 @@ ion_bands_init (mode, freqmin, freqmax, band)
     xfreq[8] = 3.162e18;
     xfreq[9] = 1.2e19;          //This is the highest frequency defined in our ionization data
     xfreq[10] = freqmax;
+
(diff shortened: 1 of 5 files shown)


jhmatthews:
Regression tests look good for this - I checked that nothing changes when you use TDE and AGN banding with matrix_pow. Things do change when you use TSTAR or CV banding with matrix_pow (as I expected from my changes), but overall the results look better, so I'm merging. The reporting is also helpful, if rudimentary.

PR #1118, opened 2024-10-28
Cleanup

kslong:
Clean up of regression tests, and removal of old version of regression tests


jhmatthews:
Changed the base to dev here, which should be the default, and merging in.

PR #1128, opened 2024-11-13
Handling of low frequency photons in macro mode

kslong:
* Changes the treatment of low frequency photons generated by macro-atoms so they are not immediately absorbed
* Adds counters to plasma cells to count how many scatters of different types take place in a cell in macro mode
* Some doxygen related updates
* Some improvements to plot_wind.py

### DIFF
--- a/py_progs/plot_wind.py
+++ b/py_progs/plot_wind.py
@@ -140,6 +140,12 @@ def get_data(filename='fiducial_agn_master.txt', var='t_r',grid='ij',inwind='',s
     This routine reads and scales the data from a single variable that is read from an ascii table 
     representation of one or more of the parameters in a windsave file
 
+    The grid obptions:
+
+    ij
+    log
+    lin
+
     '''
 
     try:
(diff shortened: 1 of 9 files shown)


jhmatthews:
It looks to me like there should also be a change in the function ```f_kpkt_emit_accelerate``` - currently we have a line there where we calculate the total free-free emission again so as to normalise the contribution of free-free to the k->r emissivity. I think as it is in this PR it will lead to the wrong normalisation in the final spectrum for the free-free component. One way of testing this would be to check that a free-free dominated spectrum looks like the same in the ionization and spectral cycles if a wide band is chosen, although it might be tricky to find the right model.

kslong:
@jhmatthews   I guess I don't know enough to understand why this would be a concern, as you and Sturat are the experts on acceleration in the macro atom case.

However, fhe caller graph for this indicates it is only used to define photons over a limited range, as you mention in your comment, for spectral cycles.  During ionization cycles, no "wind" photons are generated so this does not come in play.

But we have always for spectral cycles limited the spectral range by a much larger amount.  An then  any photons, regardless of how they are originated, if the interact and end up outside the range then are effectively eliminated.

I can imagine this would be a problem if we used too narrow a wavelength range in the ionization cycles, and we used information from the ionization cycles to "correct" the emissivities in the spectral cycles, but that is a fundamental problem.  

image

jhmatthews:
I've made a small change to macro_accelerate to get it to work with the lofreq_ff limit. 

Note that the commented out /redundant code is still in there at the moment and could be cleared out once we're happy everything is stable. 

Additionally, note for users that this PR increases MAXSCAT to 20000.

Issue #1126, opened 2024-11-08
communicate_plasma is undocumented


kslong:
Communicate_plasma (and to a lessor extent the other communicate routines) is (are) quite complex, and need good documentation to allow anyone except @Edward-RSE  or @jhmatthews to modify anything in the Plasma structure, especially with regard to understanding changes to things like buffer size. 

An example of this is the lack of explanation for a line like:

```
  const int comm_buffer_size = calculate_comm_buffer_size (1 + n_cells_max * (1 + 20 + nphot_total + nions + NXBANDS + 2 * N_PHOT_PROC),
                                                           n_cells_max * (71 + 11 * nions + nlte_levels + 2 * nphot_total + n_inner_tot +
                                                                          11 * NXBANDS + NBINS_IN_CELL_SPEC + 6 * NFLUX_ANGLES +
                                                                          N_DMO_DT_DIRECTIONS + 12 * NFORCE_DIRECTIONS));
```
where, to point to one specific issue, what does 71 refer to.

This is relevant to me currently, as I would like to add, some variables to track, the number of ff and bf scatters in each cell to better understand why some of the Pluto models seem to be exceedingly slow.

(As an aside, also odd from my perspective, if you look at communicate_plasma, is there are places where nscat_es is "communicated", but nscat_nres is not. )

Instructions are needed to explain what one should do to make at least  "minor" modifications, like adding an integer or double or an array to the structure, either as part of the doxygen comments in the file, or in an rst file for the sphinx documentation.

Edward-RSE:
I can add some inline comments to explain the code better, and I'll also update the Doxygen comments for the functions to explain how to add another variable to the communication.

To clarify what `71` is, it's the number of *single* doubles which are communicated for each cell. This corresponds to the number of `MPI_Pack`'s which looks like this:

```c
MPI_Pack (&cell->vol, 1, MPI_DOUBLE, comm_buffer, comm_buffer_size, &position, MPI_COMM_WORLD);
```

Then, for example, `11 * nions` corresponds to communicating 11 arrays which each have `nions` number of elements. This is the same approach we used prior to the communication refactor, and I believe it’s the most efficient approach we can take -- but that's not to say we can't improve its clarity or documentation. Using a derived type to avoid manually counting variables to determine the buffer size is not feasible, due to the pointers in the `PlasmaPtr` structure meaning `plamamain` is not contiguous in memory.

[...]

PR #1129, opened 2024-11-15
Improve MPI communication documentation

Edward-RSE:
This PR is addressing issue #1126, which states the documentation for MPI communication is not sufficient for the complexity of the code.

To address this I have written a new page in the Sphinx documentation which breaks down the communication code in detail. This page documents how the standard communication pattern works and how the size of the comm buffer is calculated (with simple code examples). It also includes a section on how to add a new variable to an existing communication.

I opted to create a Sphinx page instead of adding to the Doxygen comments for each function, to keep that comment block from becoming too unwieldily and long. In each comment block, I  reference the new Sphinx page for anyone who wants more details.

(I also noticed that the malloc's in a lot of the functions were unchecked, so I have added checks to make sure memory is allocated correctly)

### DIFF
--- a/docs/sphinx/source/developer.rst
+++ b/docs/sphinx/source/developer.rst
@@ -7,5 +7,6 @@ This page contains documentation intended for developers.
     :glob:
 
     developer/programmer_notes
+    developer/mpi_comms
     developer/cuda
     developer/tests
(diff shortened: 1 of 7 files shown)


kslong:
I think that is a good way to have dealt with this issue, including having a reference in the .c file to the documentation on sphinx.  I am happy for this to be merged.

PR #1135, opened 2025-01-31
don't call call qdisk_init if we are restarting

### LINKED ISSUE #1134: Information about the temperature of irradiated disk (when using thermalized.rerad) is lost when restarting a run
When using the thermalized.rerad mode the temperature of the disk can increase because of photons hitting the disk and being absorbed.  When restarting a run that uses the thermalized.rerad mode, the temperature of this irradiated disk at the previous cycle is lost. In the plot attached, I did 50 ionization cycles in a run with a lot of irradiation of the disk before restarting the run for 1 more cycle. As can be seen, the disk before restarting is much more luminous and hot (because of irradiation) than after restarting (because it started back from the input disk SED without any irradiation). I attached the .pf, .pluto, (after the first 50 cycles) files as well as the .log_spec_tot at the 50th and 51st cycles. I cannot attached the .wind_save file because it's too big.

A simple fix would be to save the information about the temperature of the disk in .wind_save file so that it can be used at initialization for restarts. 

after_before_restart_comparison
[example.zip](https://github.com/user-attachments/files/18394396/example.zip)

jhmatthews:
This is designed to fix problems with the disk spectrum resetting when you restart - this should not happen. The solution is to only call qdisk_init if geo.run_type != RUN_TYPE_RESTART. This should help fix #1134

### DIFF
--- a/source/sirocco.c
+++ b/source/sirocco.c
@@ -645,7 +645,11 @@ main (argc, argv)
 
 
   disk_init (geo.disk_rad_min, geo.disk_rad_max, geo.mstar, geo.disk_mdot, freqmin, freqmax, 0, &geo.f_disk);
-  qdisk_init (geo.disk_rad_min, geo.disk_rad_max, geo.mstar, geo.disk_mdot);
+
+  /* we should not call qdisk_init if we are restarting, see #1134 */
+  if (geo.run_type != RUN_TYPE_RESTART)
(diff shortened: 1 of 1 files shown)


jhmatthews:
Together with Nico we've verified that this gives the desired behaviour (i.e, that you retain the qdisk properties from the wind_save file when using restarts such that the irradiated disk spectrum appears straight away in the restarted run), so I'm merging this into dev.

Issue #1137, opened 2025-03-01
Failure to link data directories


kslong:
For the models associated with the CMFGen tests, sirocco is not setting up the links to the data directories properly, even though it works well for the standard regression tests.  The CMFGen test models are a bit unusual in that they read non-standard abundance files, as well as a user defined continuum, but still this should NOT be hapening.

kslong:
Verified that this is also happening on Mac. The error that is supposed to indicate an inability to run this is not occurring

The creating of the links is done in Setup.c (around line 936) is simply incorrect, and assumes that if the masterfile can be read, then the data directories will have been linked.  Often, we modify the master file locally, but we expect some of the detailed files to be do in data or xdata.

wind_read also has a Setup_Sorocco_Dir command which is also odd.

It looks there would be no harm in simply always running the Setup_Sorocco_Dir, unless someone has created a private version, of say data, and linked it.  

Setup_Sirocco_Dir removes and reestablishes symbolic links to data, etc, but does not rm actual directories (but would instead throw and error because the directory is not a link).

kslong:
I have not seen this issue in a while.  I propose to close it until it reappears.

PR #1138, opened 2025-03-04
Bug1137 data

kslong:
Addresses #1137 where the data directories were not being linked if one had a master file in the local directory, and had not previously run Sirocco in that directory

### DIFF
--- a/bin/Setup_Sirocco_Dir
+++ b/bin/Setup_Sirocco_Dir
@@ -1,48 +1,39 @@
 #! /bin/bash
 
-# Please make sure your $SIROCCO path is set up to use this script
+
+
+# please make sure your $SIROCCO path is setup to use this script
 top=$SIROCCO
 xdata=xdata
 xmod=xmod
 zdata=zdata
 
-# Function to check if a directory exists, is NOT a symlink, and contains files
-check_and_warn() {
-    if [ -d "$1" ] && [ ! -L "$1" ] && [ "$(ls -A "$1" 2>/dev/null)" ]; then
-        echo "Warning: $1 is a real directory and contains files. It will not be relinked."
-        return 1
-    fi
-    return 0
-}
-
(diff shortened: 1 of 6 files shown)


jhmatthews:
This currently fails because it's unable to run Setup_Sirocco_Dir, but I'm not sure why this is. It looks like it's actually struggling to find the script in the github workflow, but this was working before, so it's odd. 

Note that to diagnose this you can click on the failed test above and see what commands were run before the fail. You can then either run these commands locally, or I guess one could use something like act: https://github.com/nektos/act

kslong:
I did change that script, but I did not change it's name, and the only thing it does, is avoid creating a directory if it is already there.  It prints a message if it found a "true" directory and not a link.

Is there a yaml file or something that describes what is supposed to happen in the workfile.

jhmatthews:
Yeah, indeed, that's why I think it's odd

The yaml file is in .github/workflows/, or can be viewed through the test itself on github actions.

kslong:
OK, I understand why this is failing the test.

The previous version of the code only ran Setup_Sirocco_Dir if the master file could not be found.  This was never the case in the continuous integration tests because we run Setup_Sirocco_Dir as part of the setup before running sirroco.j

In the new version of the code, we are always running Setup_Sirocco_Dir from sirocco itself. The problem is not dued to changes made directly to Setup_Sirocco_Dir but rather due to running it all.   

In the yaml file, you can see that when we do tests, we always include the path to the executable, but of course when we run it from inside Sirroco, the assumption is that we have the bin directory in the PATH

kslong:
@jhmatthews  or @saultyevil could one of you look into what needs to be changed in the continuous integration tests so that bin is included in the PATH when these tests are run.  It's above my level of expertise.

kslong:
I am going to close this pull request as out of date at this point.  If the problem arises again we should reinvestigate.

PR #1150, opened 2025-07-15
Save photons

### LINKED ISSUE #1149: The diagnostic file root.ext.txt can be  corrupted when there are multiple threads
The diagnostic routine save_phots writes to root.ext.txt regardless of the number of threads,

If multiple threads are trying to write to the file at the same time, the file can become corrupted.  

Instead of a single file, we should open one file for each thread, similarly to what is done for the main diag files.

kslong:
The only substantive change here is to write the extra diagnostics files to different files for separate threads to prevent collision.  This should resolve #1149


kslong:
This wasn't the branch I wanted to merge with.

PR #1172, opened 2026-02-09
Refactoring from static arrays to dynamically allocated arrays and protective change for MPI communication with uninitialised memory (Low Risk Changes)

AustenWallis:
OPTIONAL ADDITIONS

One is setting up refactoring Topbase_phot phot_top[NLEVELS] and inner_cross[N_INNER*NIONS] from static arrays to dynamically allocated arrays (in atomicdata_init.c, atomic_extern_init.c and atomic.h). The Valgrind (I think this is where I identified it from) has these two arrays as very large static arrays (dominate over everything else). My thought was that these were too large heap allocations that were overflowing memory (not the case). But this change does ensure we aren’t overly “wasting” resources.

The other is a protective change where the code zeros unused band slots from bands_spec.c / wind_updates2d.c, in case the MPI communicated uninitialised memory. This was when investigating the error was likely due to the macro atom banding.

### DIFF
--- a/source/atomic.h
+++ b/source/atomic.h
@@ -368,10 +368,10 @@ typedef struct topbase_phot
   double f, log_f, sigma, log_sigma;            /**< last freq, last x-section and log versions*/
 } Topbase_phot, *TopPhotPtr;
 
-extern Topbase_phot phot_top[NLEVELS];
+extern Topbase_phot *phot_top;
 extern TopPhotPtr phot_top_ptr[NLEVELS];       /**<  Pointers to phot_top in threshold frequency order - this */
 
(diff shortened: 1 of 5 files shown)


kslong:
It is interesting you are mentioning this.  It have been thinking about creating a hybrid model for parallel operations, basically so we could have more active cells,  and Claude noted this was one of the things we should do first.  The current limits are set for CNO macro atoms I believe.  Claude thinks with shared memory we could keep the program significantly smaller, but was unclear on how this works across nodes.

jhmatthews:
This looks good Austen, merging.

Issue #1103, opened 2024-10-08
LTE_Tr (and LTE_Te) does not calcuate ionization for macro atoms in LTE


kslong:
LTE is used to initialize a Python cycle, but in LTE mode even in subsequent cycles macro atoms are ionized based on estimators due to photon passages. As a result, macro atoms are not held in LTE for the LTE_Tr and LTE_te modes.  This should not be the case.  The LTE modes should force all ions, macro or otherwise, to LTE based on Tr or Te.

(Note that it is unclear to me how some of the other "simple" modes operate)>

jhmatthews:
Knox, is this closed by the referenced PR above? #1173 ?

kslong:
Yes, @jhmatthews  This should be resolved now, so I am closing the issue.

PR #1173, opened 2026-02-10
Lte iterate

kslong:
This creates a new ionization mode lte_iterate to handle the case where all of the ion populations and levels are in lte, even though the radiation field need not me.  It also has a new procedure for matching heating and cooling, where the heating as well as the cooling is estimated.  What is done is to scale the measured heating by the ion and level densities for the new temperature.

Note that I've merged the latest dev into this, so I don't immediately understand merge conflicts.


kslong:
OK, I have resolved the changes.

kslong:
This should address among other items issue #1103

jhmatthews:
A couple of quick questions Knox:
* I guess regression tests are probably unaltered by these changes given how self-contained they are? 
* I don't really understand the "scaling" aspect, or what the motivation is? Just since normally our approach is to fix the heating, correct? 
* Some of this is because I find some of the comments pretty hard to parse - it's an LLM thing, I think. Examples would be "damp the ionization-opacity feedback loop." and "MC-phase". I find LLM comments can help but tend to be on the vague/jargon-heavy side and lack nuance.

kslong:
*
The regression tests are not affected,  nor would be any calculations involving  the non-diagnostic modes
  *
Yes, in our current approach, we never alter the heating, and only modify the temperature (and ionization) to match the predetermined heating.  In lte_iterate, we modify the heating as well, scaling the heating where possible to new ion densities.  This is “rigorous” if the cell is optically then, but of not if it optically thick.  However, as per all out iteration routines the goal is to get to a final answer in which the heating and cooling should be matched at the outset, and so no changes are needed.  There are improvements that could be made in principle, and possibly in fact, but some are limited by the quantities we capture during photoionization
  *
I have updated the description in the rst. Hopefully, it is now acceptable.  If it is unclear then. We should arrange a time to chat.

Aside: I had hoped this mode might not exhibit the stickiness we have been seeing in the very dense case. However, it does not seem to help.

From: James Matthews ***@***.***>
Date: Wednesday, February 18, 2026 at 4:07 AM
To: sirocco-rt/sirocco ***@***.***>
Cc: Knox Long ***@***.***>, Author ***@***.***>
Subject: Re: [sirocco-rt/sirocco] Lte iterate (PR #1173)

[https://avatars.githubusercontent.com/u/3007249?s=20&v=4]jhmatthews left a comment (sirocco-rt/sirocco#1173)<https://urldefense.com/v3/__https://github.com/sirocco-rt/sirocco/pull/1173*issuecomment-3919875078__;Iw!!CrWY41Z8OgsX0i-WU-0LuAcUu2o!39LaSfR32rr_7iZ9-7B10obabXtH-mvYNt7LDMeck-2atrfD0En18ZWKQpumHC8BNqOLqKfklIau4bYxI9oLqw$>

A couple of quick questions Knox:

  *   I guess regression tests are probably unaltered by these changes given how self-contained they are?
  *   I don't really understand the "scaling" aspect, or what the motivation is? Just since normally our approach is to fix the heating, correct?
  *   Some of this is because I find some of the comments pretty hard to parse - it's an LLM thing, I think. Examples would be "damp the ionization-opacity feedback loop." and "MC-phase". I find LLM comments can help but tend to be on the vague/jargon-heavy side and lack nuance.

jhmatthews:
Thanks, that makes sense. I'm happy to merge this then.

PR #1180, opened 2026-04-14
Memory

kslong:
Switch to mixed memory model for Sirocco.  In order to save memory.  This save 30-50% in memory for typical macro atom cases.  

There is no change to spectra produced by the standard regression tests.


jhmatthews:
This seems to fail the continuous integration tests at the moment, specifically the unit tests.

kslong:
I will look into this.

kslong:
At this point, this branch reproduces the regression standard regression tests exactly and passes the unit tests on both linux and Mac's.  Updates have been made to the sphinx documentation to explain the new memory model.  It would be useful if @aminmosallanezhad  would verify that the interaction to Pluto is accurately preserved, and or to add a definite test for others to verify this.

jhmatthews:
Ok so I've somewhat reviewed this PR. Obviously this is a big improvement to memory, but was hard work to review. Here are a few issues it throws up - not saying all of these prevent merge, but noting nonetheless:
* `Claude.md` has a few errors in, and seems too verbose? Given this guidance from Claude code about LLM degradation and how much contextual information an LLM can "hold onto": https://code.claude.com/docs/en/best-practices
* Do we have a summary of tests somewhere? I don't think running the regression tests with the same number of cores is sufficient if the change is to memory and parallelisation, and we would probably want also checks for memory leaks and so on? 
* Quite a lot of the commits use Claude Sonnet, I would have thought we want to use a more advanced model for complex coding tasks. 

Two things I'd like us to move towards in future: 
* Minimal AI-produced documentation or commit messages, particularly when it comes to nuanced numerical or physics aspects. Perhaps this is already the case, but I found the docs and commit messages hard to read and technical. 
* Smaller self-contained PRs into a feature branch so that changes can be easily reverted if necessary, and to make review easier.

kslong:
All of these comments are relevant for future work, as we move forward. But I hope you will simiply merge it now @jhmatthews.  The reason Sonnet was used, is because that was the model at the time, which was more than 4 months ago.  This passes all the unit tests, and all of the regression tests.

kslong:
A couple of other (possibly) more useful comments:

* This is really one change even though there are a lot of commits.  A way to simplify this should be to use **squash and merge** rather than a simple merge.  This was genarally my approach in the past.
* Also, while it is true that Claude is verbose on it's commit comments, it is also true that it does not share memories between machines.  It often reads the commit comments to understand what it has done.  Having this information in this case may have been important, because the details of the way Apple and Ubuntu handle shared memory differ.  A good bit of the back and forth here was to get this part correct

jhmatthews:
OK, points taken, but I still think we need a sustainable strategy for code review and maintainability. Squash and merge works ok for combining commit messages or changes that went back and forth, but the fundamental nature of the diff is the same - this has 7,000 line changes and 140 file changes which makes it a big outlier in our pull requests, and made review very difficult. It's more the volume of the change that is the problem rather than the number of commits.

jhmatthews:
On the tests -- have we done anything beyond the standard regression tests or anything to do with memory? i.e. how are we protected against problems specific to certain numbers of cores and/or memory leaks/overflow type things?

kslong:
It has been tested in single processor mode and with varying number of cores on both linux and macs, and I believe @aminmosallanezhad  has done so as well.

